PR #4, opened 2015-01-19
Remove package data and fix bug in proper motion correction

taldcroft:
This changes where the miniagasc.h5 file is located to better support conda skare.

It also fixes an apparent bug in the proper motion correction where the RA term is not divided by cos(Dec).  As I understand the PM_RA value is in milliarcsec / year, i.e. a normal angular distance.  It looks like the same issue is in Ska-Agasc.

This PR also adds *_PMCORR values to the output of `agasc.get_star()`.

It's not ideal to wrap up these unrelated fixes, but I want the PM fix in conda and I don't see a good reason to not put the conda-driven changes into Ska flight.

@jeanconn

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,10 @@
 mp_get_agasc_bug.png
 miniagasc.h5
-agasc/ra_dec.npy
+ra_dec.npy
 agasc1p6.h5
 doc/_*
 build
+data/
+dist/
+MANIFEST
+parsetab.py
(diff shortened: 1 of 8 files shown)


jeanconn:
Do you want to stick in that tiny EPOCH fix in this PR too?

taldcroft:
That would make sense.

taldcroft:
@jeanconn - do you recall any reason that we shouldn't merge this and install (at least in Skare 0.18?)

jeanconn:
I don't recall any reason not to run with this.

PR #7, opened 2016-12-08
Do position filtering after proper-motion correction is applied

### LINKED ISSUE #3: get_agasc_cone searches based on uncorrected positions
It looks like the largest proper motion correction in the catalog (now) is only about 0.04 degrees.  If we were to expand the search radius by the largest proper motion, we could then filter on corrected positions and not worry about this in other tools.

taldcroft:
Closes #3.


jeanconn:
:+1:  This looks good to me.  The largest offset is bigger than the 0.04 deg I quoted in #3 2 years ago but certainly smaller than the 0.1 deg you've used in the fetch pad.

taldcroft:
And for a lazy Friday project, I modernized and improved the tests.

PR #8, opened 2017-01-03
Python 3 compatibility

taldcroft:
Rename of `timeit.py` was due to some oddity when running `python setup.py test` ==> `IPython` was for some reason importing a module `timeit` and got this one by accident.


agasc/agasc.py
```diff
@@ -12,6 +12,8 @@
 
 DATA_ROOT = ska_path('data', 'agasc')
 
+tables_open_file = getattr(tables, 'open_file', None) or tables.openFile
```
jeanconn: Are the tables method changes a Python 3 issue, a PyTables API issue, or both (need newer PyTables for Python 3)?

taldcroft: This is a PyTables 2.x vs PyTables 3.x issue.  They made a major API change with 3.x.  I was thinking about a more general compatibility layer, but there are few enough instances of hdf5 access in our production code that it probably is not worth it.

taldcroft: PyTables 3.x is required for conda install in Python 3.

jeanconn:
:+1:

Issue #5, opened 2015-12-09
Can get_agasc_cone work with an agasc_file option at this point?


jeanconn:
It looks like the property reading assumes that ra_dec.npy will be in DATA_ROOT so I think that needs to be directly overridden if you want to specify/use a different agasc file in get_agasc_cone.

jeanconn:
Closed by #11

PR #11, opened 2017-04-20
Remove use of ra_dec.npy

jeanconn:
Make the position index (RA_DECS) directly from the supplied agasc h5 file instead of loading the RA Dec values from a separate file (ra_dec.npy).  This change lets the options for using a separate/non-standard agasc_file actually work and appears less than 1% slower in some %timeit testing.


jeanconn:
Closes #5

agasc/agasc.py
```diff
@@ -132,6 +133,10 @@ def get_agasc_cone(ra, dec, radius=1.5, date=None, agasc_file=None,
     # Possibly expand initial radius to allow for slop due proper motion
     rad_pm = radius + (0.1 if pm_filter else 0.0)
 
+    global RA_DECS
+    if RA_DECS is None:
```
jeanconn: I suppose I want to reinitialize this index if RA_DECS is None *or* if the query wants to use a different agasc_file than the last one query.  Hmph.

taldcroft:
What about import time?

jeanconn:
How do you generally measure that?  It looks to me like once miniagasc.h5 is .. hot in cache both imports are the same.

taldcroft:
What about cold cache case?  I think you can do `%time import agasc` no?

jeanconn:
Not sure about which caches exist to do that right (netapp vs local machine?) or how to know something isn't in the cache.

taldcroft:
One can be confident a file is not in cache by copying it.  I realized the import won't be affected by the first query would be.

Anyway I don't have time to recreate exactly why I split this out but I see the issue it creates. So I'm ok with this if you are very confident it won't break anything.

Btw what's your use case driving this?

jeanconn:
> Btw what's your use case driving this?

I was just trying to read the full agasc to get some numbers from Python on the super faint stars that are being considered as spoilers in SAUSAGE and remembered this was an open low-priority issue from me.

jeanconn:
Regarding `%time import agasc`, I thought that the ra_decs.npy file wasn't read until the first use of get_agasc_cone, so I don't think my changes should make much difference there.

Regarding `%time` in general, here's a single bench of old vs new:
```
In [3]: os.environ['SKA'] = '/data.fido/temp/agasc'
In [4]: %time import agasc
CPU times: user 1.53 s, sys: 744 ms, total: 2.27 s
Wall time: 22.3 s
In [5]: %time agasc.get_agasc_cone(10, 20, .1)
CPU times: user 93 ms, sys: 270 ms, total: 363 ms
Wall time: 2.73 s
```
New
```
In [1]: %time import agasc
CPU times: user 1.52 s, sys: 550 ms, total: 2.07 s
Wall time: 8.07 s
In [2]: %time agasc.get_agasc_cone(10, 20, .1, agasc_file='/data.fido/temp/agasc1/data/agasc/miniagasc.h5')
CPU times: user 107 ms, sys: 248 ms, total: 355 ms
Wall time: 2.5 s
```

agasc/agasc.py
```diff
@@ -132,6 +137,10 @@ def get_agasc_cone(ra, dec, radius=1.5, date=None, agasc_file=None,
     # Possibly expand initial radius to allow for slop due proper motion
     rad_pm = radius + (0.1 if pm_filter else 0.0)
 
+    global RA_DECS
```
taldcroft: Since this is a core package, let's make this a little more idiomatic and encapsulated.  

Revert the original `RA_DECS = RaDec()`.  Then add a setter to that class (`@agasc_file.setter`) which allows setting the `agasc_file` and also clears the cache (by `del self.ra; del self.dec`) (but only if the new agasc_file is different from the current one).  Then you need `@ra.deleter` and `@dec.deleter`, which delete the `_ra` or `_dec` attr if it exists.  And finally:
```
if agasc_file is not None:
    orig_agasc_file = RA_DECS.agasc_file
    RA_DECS.agasc_file = agasc_file

...

if agasc_file is not None:
    RA_DECS.agasc_file = orig_agasc_file
```
This will be slow because it initializer `RA_DECS` for each call with a custom `agasc_file`.  But if you want to work in general with a different file you could explicitly do `RA_DECS.agasc_file = my_agasc_file` at the beginning.

jeanconn: OK, though isn't the current API more idiomatic than suggesting that the user explicitly set agasc_file if he or she wants it to use a custom file and be fast?

taldcroft:
I just don't like the unexpected side effect of permanently changing the agasc file in the current implementation.

jeanconn:
I'm a little confused by that concern.  The side effect is that the agasc_file in the RA_DECS object is changed.  But if the user makes a request with no agasc_file defined, that gets reset to the default miniagasc even if a custom agasc was used the last query.

jeanconn:
Totally onboard with removing the `global` crap, but was wondering a bit about implementation in that I thought something like
```
   RA_DECS.agasc_file = agasc_file if agasc_file is not None else AGASC_FILE
```
would match this branch's API, but if you don't like the persistent non-default RA_DECS, that's OK.

taldcroft:
So what about making the RA_DECS be a dict that caches the relevant objects keyed by absolute file name.  That should mostly accomplish the original goal of allowing for use of custom files without any unexpected side effects.  The default would always be to use the "flight" miniagasc.

jeanconn:
OK, I was hoping I could convince you that since the lazy-persistent non-default RA_DECS don't have any user-visible side effects (unless the user is already reading and manipulated RA_DECS directly) that I might get away with it.

jeanconn:
I suppose the "flight" niniagasc ra, dec columns would be about 130Mb of memory, so yes that's no big deal in context... I was thinking it was a bit larger.

jeanconn:
And, for the standard use case I had of just using a different agasc, while the object for the "flight" miniagasc would be initialized in the cache, the ra and dec values would never be read into memory in the session.

agasc/tests/test_multi_agasc_file.py
```diff
@@ -0,0 +1,40 @@
...
+
+def test_multi_agasc():
+
+    tempdir = tempfile.mkdtemp()
+
+    # Make two custom agasc files from the miniagasc, using 20 stars from
+    # around the middle of the table
+    with tables_open_file(os.path.join(DATA_ROOT, 'miniagasc.h5')) as h5:
+        middle = len(h5.root.data) / 2
+        stars1 = Table(h5.root.data[middle:middle+20])
```
jeanconn: This syntax isn't Py3 compatible and should be fixed.

agasc/tests/test_multi_agasc_file.py
```diff
@@ -0,0 +1,40 @@
...
+    assert len(all_stars1) == len(stars1)
+    all_stars2 = agasc.get_agasc_cone(0, 90, radius=180, agasc_file=os.path.join(tempdir, 'stars2.h5'))
+    assert len(all_stars2) == len(stars2)
+
+    # Do a position filtered search using the first star in the table as a reference and make sure
+    # we get the same star from the reference agasc
+    cone2 = agasc.get_agasc_cone(all_stars2['RA'][0], all_stars2['DEC'][0], radius=0.000001,
+                                 agasc_file=os.path.join(tempdir, 'stars2.h5'))
+    cone2_full = agasc.get_agasc_cone(all_stars2['RA'][0], all_stars2['DEC'][0], radius=0.000001)
+    assert cone2[0]['AGASC_ID'] == cone2_full['AGASC_ID']
```
taldcroft: This asymmetry of testing `cone2[0]` (a scalar) vs. `cone2_full` (array) is bothersome.  Shouldn't it be `np.all(..)` with the full arrays?  And maybe that the length of both is 1?

agasc/tests/test_multi_agasc_file.py
```diff
@@ -0,0 +1,40 @@
...
+from .. import agasc
+
+DATA_ROOT = ska_path('data', 'agasc')
+tables_open_file = getattr(tables, 'open_file', None) or tables.openFile
+
+def test_multi_agasc():
+
+    tempdir = tempfile.mkdtemp()
+
+    # Make two custom agasc files from the miniagasc, using 20 stars from
```
taldcroft: Is the second AGASC file really adding any value to the testing?  It just confuses me because I am looking for what is different in the cases.

jeanconn: The second agasc file is confirming that you can switch back and forth between custom agasc files and it doesn't break anything and that whatever you are doing for the caching (either persistence or explicit caching) is appropriately set.

agasc/tests/test_multi_agasc_file.py
jeanconn: The length of both is 1 so this is a no-impact bug in my test code.  Will fix.

agasc/tests/test_multi_agasc_file.py
jeanconn: I suppose that switching "back an forth" would be more explicit if I did a get_cone_stars from one then the other, and then did the final star position check from the first one again instead of from the second one.

taldcroft: That would help, but more to the point a comment to that effect.

agasc/tests/test_multi_agasc_file.py
taldcroft: The length of both is *supposed* to be 1, but if cone2 ended up with 2 entries the test could erroneously pass.

agasc/tests/test_multi_agasc_file.py
jeanconn: Yes, I picked up on that too. Thanks!

agasc/tests/test_multi_agasc_file.py
```diff
@@ -0,0 +1,40 @@
...
+    # around the middle of the table
+    with tables_open_file(os.path.join(DATA_ROOT, 'miniagasc.h5')) as h5:
+        middle = len(h5.root.data) / 2
+        stars1 = Table(h5.root.data[middle:middle+20])
+        stars1.write(os.path.join(tempdir, 'stars1.h5'), path='data')
```
jeanconn: This part looks to need h5py on Py3 which isn't in the current environments.

agasc/tests/test_multi_agasc_file.py
jeanconn: Comments added and order of operations changed for more "back and forth".

agasc/tests/test_multi_agasc_file.py
jeanconn: The issue was with Py2 versus Py3 division which has since been fixed.

agasc/tests/test_multi_agasc_file.py
```diff
@@ -0,0 +1,49 @@
...
+        stars2.write(os.path.join(tempdir, 'stars2.h5'), path='data')
+
+    # Fetch all the stars from a custom agasc and make sure we have the right number of stars
+    # with no errors
+    all_stars2 = agasc.get_agasc_cone(0, 90, radius=180, agasc_file=os.path.join(tempdir, 'stars2.h5'))
+    assert len(all_stars2) == len(stars2)
+    # Fetch all the stars from the other custom agasc and do the same.  The point of the two files
+    # is to confirm that the caching behavior in agasc doesn't cause problems with fetches
+    all_stars1 = agasc.get_agasc_cone(0, 90, radius=180, agasc_file=os.path.join(tempdir, 'stars1.h5'))
+    assert len(all_stars1) == len(stars1)
```
taldcroft: Both of the custom files have 20 stars, so this test is not as strong as it should be.

agasc/tests/test_multi_agasc_file.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # Do a position filtered search using the first star in the table as a reference and make sure
+    # we get the same star from the reference agasc.  Do this with the stars2 file as this confirms
+    # that we can switch back and forth between files and get the correct content.
+    cone2 = agasc.get_agasc_cone(all_stars2['RA'][0], all_stars2['DEC'][0], radius=0.000001,
+                                 agasc_file=os.path.join(tempdir, 'stars2.h5'))
+    # And this is a read of the default agasc file after the custom ones so should confirm that
+    # the custom files didn't break that access.
+    cone2_full = agasc.get_agasc_cone(all_stars2['RA'][0], all_stars2['DEC'][0], radius=0.000001)
```
taldcroft: This could accidentally succeed if the `stars2.h5` file was still being used.  Better to pick an RA/Dec that is NOT in either stars1 or stars2, and use a non-tiny radius.  Then just check that the length is some fixed and expected value like 85.

jeanconn: A couple of thoughts here.  1) I was trying to set up these tests so if the default miniagasc changes the tests don't break.  2) The failure mechanism I've been checking is really that the right RaDecs are used.  If there is a mismatch between the RaDecs and the agasc_file, this current test will throw indexing errors, so it is testing that.

Maybe instead of trying NOT get stars, we could just run the test_multi tests before the regular agasc tests?  Wouldn't that be a safer way to get more coverage?

agasc/tests/test_multi_agasc_file.py
jeanconn: Good point, I forgot that I gave up on more complicated tests here and just went with the length tests...

agasc/tests/test_multi_agasc_file.py
taldcroft: The compatible way to write that is `len(h5.root.data) // 2`.

agasc/tests/test_multi_agasc_file.py
taldcroft: h5py should be added to ska3.  I'm don't recall why it isn't already there since dumping a table like this is quite useful.

agasc/tests/test_multi_agasc_file.py
jeanconn: I honestly wasn't sure if having them both make a float or having them both make an int was more compatible, but I suppose the int would have been less confusing.

taldcroft: Indexing with a float is either an error or raises a deprecation warning.  So there is a clear winner.

jeanconn:
​Well, just after this I was casting, but yes, I had already taken your
point.​

Issue #13, opened 2017-07-19
Proper-motion test wasn't future proof


jeanconn:
It looks to me like `test_proper_motion` assumed that the current runtime was sufficiently close to`2017:001` for this to be true:

```
star = agasc.get_star(1180612288)  # Uses current time to determine RA_CORR / PM_CORR
radius = 2.0 / 3600  # 5 arcsec
stars = agasc.get_agasc_cone(star['RA_PMCORR'], star['DEC_PMCORR'], radius, date='2017:001')
assert len(stars) == 1
```
It looks to me that if this should be a static test, that it should use:
```
star = agasc.get_star(1180612288, date='2017:001')
```

jeanconn:
Closed by #14

PR #14, opened 2017-07-19
Update test to use fixed time for PM corr. for ref. star

jeanconn:
Update test to use fixed time for PM correction for reference star.


jeanconn:
Closes #13

PR #15, opened 2017-08-09
Only calculate times for proper motion for stars that need it

jeanconn:
Only calculate times for proper motion for stars that need it.

This reduces the number of DateTime fracyear conversions from thousands to, basically 1, for most fields.

### DIFF
--- a/agasc/agasc.py
+++ b/agasc/agasc.py
@@ -99,12 +99,14 @@ def add_pmcorr_columns(stars, date):
     # The dyear for proper motion is only relevant for stars that have defined proper motion
     # so set to zero for all stars by default
     dyear = np.zeros(len(stars))
+    has_pm = (stars['PM_DEC'] != -9999) | (stars['PM_RA'] != -9999)
     # For most of them, the epoch is fixed at 2000, so we don't need N calls to DateTime to
     # figure that out
-    ok = ((stars['PM_DEC'] != -9999) | (stars['PM_RA'] != -9999)) & (stars['EPOCH'] == 2000.0)
+    epoch_is_2000 = (stars['EPOCH'] == 2000.0)
+    ok = has_pm & epoch_is_2000
     dyear[ok] = (DateTime(date) - DateTime(2000, format='frac_year')) / 365.25
-    # For any other EPOCHs that have a proper motion correction, do them individually
-    ok = ((stars['PM_DEC'] != -9999) | (stars['PM_RA'] != -9999)) & (stars['EPOCH'] != 2000.0)
+    # For stars with proper motion correction but epoch != 2000, calculate individually.
+    ok = has_pm & ~epoch_is_2000
     dyear[ok] = (DateTime(date) - DateTime(stars[ok]['EPOCH'], format='frac_year')) / 365.25
     pm_to_degrees = dyear / (3600. * 1000.)
 
--- a/agasc/agasc.py
+++ b/agasc/agasc.py
@@ -107,7 +107,7 @@ def add_pmcorr_columns(stars, date):
     dyear[ok] = (DateTime(date) - DateTime(2000, format='frac_year')) / 365.25
     # For stars with proper motion correction but epoch != 2000, calculate individually.
     ok = has_pm & ~epoch_is_2000
-    dyear[ok] = (DateTime(date) - DateTime(stars[ok]['EPOCH'], format='frac_year')) / 365.25
+    dyear[ok] = (DateTime(date) - DateTime(stars['EPOCH'][ok], format='frac_year')) / 365.25
     pm_to_degrees = dyear / (3600. * 1000.)
 
     dec_pmcorr = np.where(stars['PM_DEC'] != -9999,
--- a/agasc/agasc.py
+++ b/agasc/agasc.py
@@ -96,7 +96,7 @@ def add_pmcorr_columns(stars, date):
     # field to degrees.  The AGASC PM is specified in milliarcsecs / year, so this
     # is dyear * (degrees / milliarcsec)
 
-    # The dyear for proper motion is only relevent for stars that have not -9999 proper motion
+    # The dyear for proper motion is only relevant for stars that have defined proper motion
     # so set to zero for all stars by default
     dyear = np.zeros(len(stars))
     # For most of them, the epoch is fixed at 2000, so we don't need N calls to DateTime to
--- a/agasc/agasc.py
+++ b/agasc/agasc.py
@@ -95,8 +95,17 @@ def add_pmcorr_columns(stars, date):
     # Compute the multiplicative factor to convert from the AGASC proper motion
     # field to degrees.  The AGASC PM is specified in milliarcsecs / year, so this
     # is dyear * (degrees / milliarcsec)
-    agasc_equinox = DateTime(stars['EPOCH'], format='frac_year')
-    dyear = (DateTime(date) - agasc_equinox) / 365.25
+
+    # The dyear for proper motion is only relevent for stars that have not -9999 proper motion
+    # so set to zero for all stars by default
+    dyear = np.zeros(len(stars))
+    # For most of them, the epoch is fixed at 2000, so we don't need N calls to DateTime to
+    # figure that out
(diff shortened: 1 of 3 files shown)


agasc/agasc.py
```diff
@@ -95,8 +95,17 @@ def add_pmcorr_columns(stars, date):
...
+
+    # The dyear for proper motion is only relevent for stars that have not -9999 proper motion
+    # so set to zero for all stars by default
+    dyear = np.zeros(len(stars))
+    # For most of them, the epoch is fixed at 2000, so we don't need N calls to DateTime to
+    # figure that out
+    ok = ((stars['PM_DEC'] != -9999) | (stars['PM_RA'] != -9999)) & (stars['EPOCH'] == 2000.0)
+    dyear[ok] = (DateTime(date) - DateTime(2000, format='frac_year')) / 365.25
+    # For any other EPOCHs that have a proper motion correction, do them individually
+    ok = ((stars['PM_DEC'] != -9999) | (stars['PM_RA'] != -9999)) & (stars['EPOCH'] != 2000.0)
```
jeanconn: Not sure what to name this tiny mask instead of re-using "ok"

agasc/agasc.py
```diff
@@ -95,8 +95,17 @@ def add_pmcorr_columns(stars, date):
     # Compute the multiplicative factor to convert from the AGASC proper motion
     # field to degrees.  The AGASC PM is specified in milliarcsecs / year, so this
     # is dyear * (degrees / milliarcsec)
-    agasc_equinox = DateTime(stars['EPOCH'], format='frac_year')
-    dyear = (DateTime(date) - agasc_equinox) / 365.25
+
+    # The dyear for proper motion is only relevent for stars that have not -9999 proper motion
```
jeanconn: Typo in the comment.

jeanconn:
In the one field I was using in testing, this improved get_agasc_cone speed by a good bit

Old
```
In [2]: %timeit agasc.get_agasc_cone(10, 20, radius=3)
767 ms ± 3.79 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
New
```
In [2]: %timeit agasc.get_agasc_cone(10, 20, radius=3)
101 ms ± 3.64 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

agasc/agasc.py
taldcroft: `ok` is pretty much always OK in this type of situation, even for re-use in close proximity.

taldcroft: But another rule is DRY.  This is easier to read and faster.
```
has_pm = (stars['PM_DEC'] != -9999) | (stars['PM_RA'] != -9999)
epoch_is_2000 = (stars['EPOCH'] == 2000.0)
ok = has_pm & epoch_is_2000
...
ok = has_pm & ~epoch_is_2000
...
```

jeanconn: Cool.  Thanks!.  I wasn't sure about DRY vs "do not use one-offs"... but I suppose I at least had a two-off .

agasc/agasc.py
```diff
@@ -95,8 +95,17 @@ def add_pmcorr_columns(stars, date):
...
+    # so set to zero for all stars by default
+    dyear = np.zeros(len(stars))
+    # For most of them, the epoch is fixed at 2000, so we don't need N calls to DateTime to
+    # figure that out
+    ok = ((stars['PM_DEC'] != -9999) | (stars['PM_RA'] != -9999)) & (stars['EPOCH'] == 2000.0)
+    dyear[ok] = (DateTime(date) - DateTime(2000, format='frac_year')) / 365.25
+    # For any other EPOCHs that have a proper motion correction, do them individually
+    ok = ((stars['PM_DEC'] != -9999) | (stars['PM_RA'] != -9999)) & (stars['EPOCH'] != 2000.0)
+    dyear[ok] = (DateTime(date) - DateTime(stars[ok]['EPOCH'], format='frac_year')) / 365.25
```
taldcroft: Another very important performance rule is never slice the table if you just slice the column.  This can save an order of magnitude or more.
```
DateTime(stars['EPOCH'][ok])
```

taldcroft:
Good catch BTW.

jeanconn:
Was still looking around in RunSnakeRun in the aca lts evaluator code and just saw a lot of calls to DateTime convert in get_agasc_cone.  This change isn't going to help much with the speed of the lts evaluator code but at least looked like something fixable.

agasc/agasc.py
```diff
@@ -95,8 +95,20 @@ def add_pmcorr_columns(stars, date):
...
+    dyear = np.zeros(len(stars))
+    # For most of them, the epoch is fixed at 2000, so we don't need N calls to DateTime to
+    # figure that out
+    has_pm = (stars['PM_DEC'] != -9999) | (stars['PM_RA'] != -9999)
+    epoch_is_2000 = (stars['EPOCH'] == 2000.0)
+    ok = has_pm & epoch_is_2000
+    dyear[ok] = (DateTime(date) - DateTime(2000, format='frac_year')) / 365.25
+    # For stars with proper motion correction but epoch != 2000, calculate individually.
+    ok = has_pm & ~epoch_is_2000
+    ok = ((stars['PM_DEC'] != -9999) | (stars['PM_RA'] != -9999)) & (stars['EPOCH'] != 2000.0)
```
jeanconn: Oops.

taldcroft:
Well this all *looks* right, but as I look at the testing I can't see anything that explicitly checks that it is giving the same values as the current flight code.  I know it is a bit of mission creep, but since this is a pretty important function for us it is probably worth putting in a regression test that covers the cases in the code including non-2000 epoch.  Ugh.

jeanconn:
Gotcha.  I was working a bit from your email comments:

> It appears there are exactly 2 out of 8320726 stars in the mini-AGASC that have a proper motion value provided and EPOCH != 2000.0:

```
In [39]: x = agasc.get_agasc_cone(30, 40, radius=180)  # get the whole thing
In [40]: bad = (x['EPOCH'] != 2000.0) & (x['PM_RA'] != -9999)
In [41]: np.sum(bad)
Out[41]: 2
```

So for standard use of the AGASC I suppose I should find those two stars and see if they work.

jeanconn:
PS I was using this agasc branch in the aca_lts_eval code to speed things up a bit, but if we notice something incorrect there's still plenty of time to rerun that analysis for record.

taldcroft:
I'm not worried about the LTS eval for this.  It's more about raising the bar a bit and making sure that new code or code changes in key packages are actually tested.

jeanconn:
The PS was mostly for me.  If I did something in this code that moved all the stars by an big incorrect chunk, I'd want to take an action to fix the only thing thus far that has used that code... and putting a note for that in this PR seemed reasonable.

jeanconn:
@taldcroft does this still need more testing?  I'm not sure if the test I added covers your request " is probably worth putting in a regression test that covers the cases in the code including non-2000 epoch".

taldcroft:
Just to make this bombproof, change the test to be a real regression test.  The current version is basically using the module code to generate the "truth".  Instead assume the current release is "truth" and hardcode the PM-corrected values from that into the test.

jeanconn:
Updated with the test now set to be a real regression test.

agasc/tests/test_agasc_2.py
```diff
@@ -213,24 +212,18 @@ def test_proper_motion():
...
-    date = '2020:001'
     star = agasc.get_star(agasc_id, date=date)
-    dyear = DateTime(date).frac_year - star['EPOCH']
-    pm_to_degrees = dyear / (3600. * 1000.)
-    pm_dec = (star['DEC_PMCORR'] - star['DEC']) / pm_to_degrees
-    assert np.round(pm_dec) == star['PM_DEC']
-    ra_scale = np.cos(np.radians(star['DEC']))
-    pm_ra = (star['RA_PMCORR'] - star['RA']) * ra_scale / pm_to_degrees
-    assert np.round(pm_ra) == star['PM_RA']
+    assert np.isclose(star['RA_PMCORR'], ra_pmcorr)
```
taldcroft: The default `rtol=1e-5` for `np.isclose` is too large here.  That implies checking that the values are close to within 0.00247 for RA of 247.  I think that `rtol=0` and `atol=1e-5` would probably be good enough.

Issue #19, opened 2018-04-03
get_agasc_cone segmentation fault


malgosias:
>from agasc import get_agasc_cone
>ra = 359.25177
>dec = 0.1305667
>get_agasc_cone(ra, dec, 0.1)

<works OK>

but

>ra = np.float16(359.25177)
>dec = np.float16(0.1305667)
>ra, dec

(359.25, 0.13062)

>get_agasc_cone(ra, dec, 0.1)
Segmentation fault (core dumped)

taldcroft:
Probably in the `numexpr` code.   Maybe something like `np.radians(ra1).astype(np.float64)` would do the trick, but haven't tested that.

jeanconn:
That does seem to be where the segfault happens.  Not sure if an astype or cast makes more sense or if this needs np.float64 or np.float32
```
In [13]: %timeit np.radians(np.float16(10)).astype(np.float64)
4.48 µs ± 10 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [14]: %timeit np.radians(np.float64(np.float16(10)))
2.26 µs ± 5.52 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [21]: %timeit np.float64(np.radians(np.float16(10)))
2.21 µs ± 4.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

```

jeanconn:
I think this was closed by #20

PR #20, opened 2018-04-04
Use astype to avoid np.float16 ras, decs causing segfault

jeanconn:
Use astype to avoid np.float16 ras, decs causing segfault #20


jeanconn:
Closes #19

PR #22, opened 2018-04-10
Only initialize cache with default agasc if available

jeanconn:
Only initialize cache with default agasc if available . This lets the module load even if there is no agasc, which is needed to install the agasc module if on non-local system.

### DIFF
--- a/agasc/agasc.py
+++ b/agasc/agasc.py
@@ -15,7 +15,11 @@ DATA_ROOT = ska_path('data', 'agasc')
 
 tables_open_file = getattr(tables, 'open_file', None) or tables.openFile
 
-DEFAULT_AGASC_FILE = os.path.join(DATA_ROOT, 'miniagasc.h5')
+if DATA_ROOT is not None and os.path.exists(DATA_ROOT):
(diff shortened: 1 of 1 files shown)


agasc/agasc.py
```diff
@@ -55,7 +59,8 @@ def read_ra_dec(self):
             # Now copy to separate ndarrays for memory efficiency
             return radecs['RA'].copy(), radecs['DEC'].copy()
 
-RA_DECS_CACHE = {DEFAULT_AGASC_FILE: RaDec(DEFAULT_AGASC_FILE)}
+if DEFAULT_AGASC_FILE is not None and os.path.exists(DEFAULT_AGASC_FILE):
```
jeanconn: Is this somewhat reasonable?  Not sure if there is a more reasonable way to just defer setting up this cache until the user actually wants to use the agasc (not just import).

taldcroft:
Have a look at `LazyDict` in `maude` or `LazyVal` in `kadi`.  These may be better solutions that allow lazy-loading of a file and a useful exception message if not possible.

taldcroft:
> which is needed to install the agasc module if on non-local system.

No longer applicable with SCM versioning.

jeanconn:
How does SCM versioning solve this?

PR #25, opened 2018-09-10
Changes to create_mini_agasc_h5 for arbitrary AGASC version

taldcroft:
Houston, we have a problem...  I decided that I wanted the 1.7 miniagasc now to start using in proseco development, so went ahead with the few patches to make this happen.  It worked for 1p6 but failed for 1p7.   In my work `agasc1p7.h5` is a link to `/proj/sot/ska/data/agasc/agasc1p7.h5`.  

It looks like there is something really different.  @malgosias - please investigate! 

```
ska-kadi$ python create_mini_agasc_h5.py --version=1p6
Reading full AGASC agasc1p6.h5 and selecting useable stars
Sorting on Dec and re-ordering
Creating miniagasc.h5 file
Appending stars to miniagasc_1p6.h5 file
Creating indexes in miniagasc.h5 file
Flush and close miniagasc.h5 file

ska-kadi$ python create_mini_agasc_h5.py --version=1p7
Reading full agasc agasc1p7.h5 and selecting useable stars
Traceback (most recent call last):
  File "create_mini_agasc_h5.py", line 24, in <module>
    h5 = tables.openFile(filename, mode='r')
  File "/proj/sot/ska/arch/x86_64-linux_CentOS-6/lib/python2.7/site-packages/tables/file.py", line 258, in openFile
    return File(filename, mode, title, rootUEP, filters, **kwargs)
  File "/proj/sot/ska/arch/x86_64-linux_CentOS-6/lib/python2.7/site-packages/tables/file.py", line 533, in __init__
    self._g_new(filename, mode, **params)
  File "hdf5Extension.pyx", line 281, in tables.hdf5Extension.File._g_new (tables/hdf5Extension.c:2180)
  File "/proj/sot/ska/arch/x86_64-linux_CentOS-6/lib/python2.7/site-packages/tables/utils.py", line 143, in checkFileAccess
    raise IOError("``%s`` does not exist" % (filename,))
IOError: ``agasc1p7.h5`` does not exist
ska-kadi$ ln -s /proj/sot/ska/data/agasc/agasc1p7.h5 ./
ska-kadi$ python create_mini_agasc_h5.py --version=1p7
Reading full agasc agasc1p7.h5 and selecting useable stars
Sorting on Dec and re-ordering
Creating miniagasc.h5 file
Appending stars to miniagasc_1p7.h5 file
Traceback (most recent call last):
  File "create_mini_agasc_h5.py", line 40, in <module>
    minitbl.append(stars)
[...]

PR #26, opened 2018-09-11
Agasc1p7

malgosias:
This is a pull request for updated create*.py scripts changed so that they work with AGASC V1.7, and V1.7 test script which:
- confirms that all the fields in all the columns match values in the input agasc1p7.h5 file (new h5 file created from Jean's fits files is in /home/malgosia/git/agasc/agasc1p7.h5, branch agasc1p7)
- confirms that there are no differences between agasc1p7.h5 file and 1.6 file in columns other than new mag_aca, mag_aca_err and rsv 1-3 columns.

These tests fail for one star entry and it is not clear to me why: the AGASC_ID is OK, but RA and DEC are slightly different, there is also a difference in COLOR2 column (and I did not check the remaining columns for this one failing entry).

What is missing is a confirmation that we do see expected differences between agasc1p7.h5 file and 1.6 in new mag_aca, mag_aca_err and rsvd 1-3 columns. But I would like to first understand the origin of this one entry that fails the above tests. Let me know if you can think of any reasons for this.

### DIFF
--- a/create_mini_agasc_h5.py
+++ b/create_mini_agasc_h5.py
@@ -5,12 +5,12 @@ import Ska.Table
 from astropy.table import Table
 
 
-example_file = '/data/agasc1p6/agasc/n0000/0001.fit'
+example_file = '/proj/sot/ska/data/agasc1p7/agasc/n0000/0001.fit'
 dtype = Ska.Table.read_fits_table(example_file).dtype
 table_desc, bo = tables.descr_from_dtype(dtype)
 
 print 'Reading full agasc and selecting useable stars'
-h5 = tables.openFile('agasc1p6.h5', mode='r')
+h5 = tables.openFile('agasc1p7.h5', mode='r')
 tbl = h5.getNode("/", 'data')
 idxs = tbl.getWhereList("(MAG_ACA - (3.0 * MAG_ACA_ERR / 100.0)) < 11.5")
 stars = tbl.readCoordinates(idxs)
@@ -30,7 +30,7 @@ np.save('ra_dec.npy', out)
 print 'Creating miniagasc.h5 file'
 minih5 = tables.openFile('miniagasc.h5', mode='w')
 minitbl = minih5.createTable('/', 'data', table_desc,
-                             title='AGASC 1.6')
+                             title='AGASC 1.7')
 print 'Appending stars to miniagasc.h5 file'
 minitbl.append(stars)
 minitbl.flush()
--- a/create_mini_agasc_h5.py
+++ b/create_mini_agasc_h5.py
@@ -1,16 +1,27 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+import argparse
+import re
+
 import numpy as np
 import tables
 import Ska.Table
 from astropy.table import Table
 
+parser = argparse.ArgumentParser()
+parser.add_argument('--version',
+                    default='1p7',
+                    help='Version (e.g. 1p6 or 1p7, default=1p7')
+args = parser.parse_args()
 
-example_file = '/proj/sot/ska/data/agasc1p7/agasc/n0000/0001.fit'
+num_version = re.sub(r'p', '.', args.version)
+
+example_file = '/proj/sot/ska/data/agasc{}/agasc/n0000/0001.fit'.format(args.version)
 dtype = Ska.Table.read_fits_table(example_file).dtype
 table_desc, bo = tables.descr_from_dtype(dtype)
 
-print 'Reading full agasc and selecting useable stars'
-h5 = tables.openFile('agasc1p7.h5', mode='r')
+filename = 'agasc{}.h5'.format(args.version)
+print 'Reading full AGASC {} and selecting useable stars'.format(filename)
+h5 = tables.openFile(filename, mode='r')
 tbl = h5.getNode("/", 'data')
 idxs = tbl.getWhereList("(MAG_ACA - (3.0 * MAG_ACA_ERR / 100.0)) < 11.5")
 stars = tbl.readCoordinates(idxs)
@@ -20,18 +31,12 @@ print 'Sorting on Dec and re-ordering'
 idx = np.argsort(stars['DEC'])
 stars = stars.take(idx)
 
-ra = np.array(stars['RA'], dtype='f8')
-dec = np.array(stars['DEC'], dtype='f8')
-
-print 'Saving lightweight index ra_dec.npy'
-out = Table([ra, dec], names=('ra', 'dec'))
-np.save('ra_dec.npy', out)
-
 print 'Creating miniagasc.h5 file'
-minih5 = tables.openFile('miniagasc.h5', mode='w')
+filename = 'miniagasc_{}.h5'.format(args.version)
+minih5 = tables.openFile(filename, mode='w')
 minitbl = minih5.createTable('/', 'data', table_desc,
-                             title='AGASC 1.7')
-print 'Appending stars to miniagasc.h5 file'
+                             title='AGASC {}'.format(num_version))
+print 'Appending stars to {} file'.format(filename)
 minitbl.append(stars)
 minitbl.flush()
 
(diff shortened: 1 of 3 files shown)


jeanconn:
What's the agasc id on the fail?

jeanconn:
It looks to me like the h5 file built from the fits file has the problem that there are two entries for star 154534513.  I don't see where this is a mistake in the code that makes the h5 files so I suspect this is a bug in AGASC 1.6 (will confirm) and we'll need to decide if AGASC 1.7 should fix the bug.

jeanconn:
I've confirmed that we have two stars that share this AGASC_ID AGASC 1.6.  It looks like you can basically crash SAUSAGE on this if you manually select the star (or stars) with the shared id, but I consider this a no-impact bug and think we should just open an issue to fix this for the next AGASC.  Is there anything else we should confirm on that course of action @taldcroft ?

test_agasc1p7.py
```diff
@@ -0,0 +1,47 @@
...
+agasc_ids = h5.root.data.col('AGASC_ID')
+agasc_ids2 = h5inp.root.data.col('AGASC_ID')
+agasc_ids3 = h51p6.root.data.col('AGASC_ID')
+
+idx = np.arange(len(agasc_ids))
+
+id_map2 = dict(zip(agasc_ids2, idx))
+id_map3 = dict(zip(agasc_ids3, idx))
+
+h5inp_id_map = [id_map2[agasc_id] for agasc_id in agasc_ids]
```
taldcroft: Why is this ID map required?  The files should have exactly the same number of rows, in exactly the same order, right?

test_agasc1p7.py
```diff
@@ -0,0 +1,47 @@
+
```
taldcroft: Add module docstring saying what this script is doing and why.

test_agasc1p7.py
```diff
@@ -0,0 +1,47 @@
...
+import numpy as np
+import tables
+
+h5 = tables.openFile('agasc1p7.h5', mode='r')
```
taldcroft: This file should be called (here) something like `agasc1p7_from_fits.h5`, and the other one should be `agasc1p7_from_jupyter.h5`.  Otherwise the heritage is very confusing.  You can do this with soft links and add a code comment describing where each file came from (first from the `create_agasc_h5.py` and second from the Jupyter notebook).

These docs won't be wasted effort since you will need to describe the file creation process in detail for the final AGASC 1.7 README (or wherever we decide that needs to go).

create_mini_agasc_h5.py
```diff
@@ -5,12 +5,12 @@
 from astropy.table import Table
 
 
-example_file = '/data/agasc1p6/agasc/n0000/0001.fit'
+example_file = '/proj/sot/ska/data/agasc1p7/agasc/n0000/0001.fit'
```
taldcroft: For this script you should use the version I created earlier that has a command line input to create either 1p6 or 1p7.

test_agasc1p7.py
```diff
@@ -0,0 +1,47 @@
+
+import numpy as np
+import tables
```
taldcroft: This test script should pass on Py2 and Py3, so also `import tables3_api`.  See https://github.com/sot/tables3_api/blob/master/README.md

taldcroft:
>  I consider this a no-impact bug and think we should just open an issue to fix this for the next AGASC. Is there anything else we should confirm?

Agreed.  I can't think of anything else.

jeanconn:
Should probably independently decide if we want to do anything with the 154534513 star in the h5 version or just leave it alone.

test_agasc1p7.py
malgosias: For some reason they don't have the same order. I'll double check this but without the ID map I was getting Assertion Errors.

create_mini_agasc_h5.py
taldcroft: I'm talking about the entire create_mini_agasc_h5.py script, not the `example_file`.  See: https://github.com/sot/agasc/blob/270f862458c10aaf5d1768e9711eb51b7f6fe083/create_mini_agasc_h5.py

From https://github.com/sot/agasc/pull/25/files.

Or am I still misunderstanding?

malgosias: Ah, OK, my mistake. I will use the version with a command line input.

malgosias:
I implemented the requested updates and added a check of all fits file comment cards, let me know if this is what you had in mind. The h5 tests still fail because of star 154534513. Should I filter on this AGASC_ID and ignore this star entry for the tests? Or will the duplicate entry be removed from the fits files?

jeanconn:
I know @malgosias is the one who needs to undestand that comment but what does "and then have separate handling for showing that one star is also ok"?  By "that one star" do you mean the duplicate star?  And what is "ok" in this context?  Or do you want a test showing that fetching any *other* single star works?

taldcroft:
@malgosias - All the tests that are documented in the comments need to pass.  The implementation of how to make this so is your challenge.  One possibility - Instead of using an ID map which will fail for the one case of a duplicate ID, could you sort by ID?  That will ensure that no entries are lost.

jeanconn:
@malgosias , regarding "Or will the duplicate entry be removed from the fits files?", my plan is to leave the duplicate in the fits files for this release.

taldcroft:
@malgosias - any progress on this?

malgosias:
With this version I use Table.read instead of tables.openFile for easier sorting (no id maps anymore).  This breaks on my computer because of memory but works on kadi. I tried using ``tables.openFile`` and then ``readSorted`` but the jupyter h5 file has no indexing.

I sort with respect to AGASC_ID and RA so that the two entries for the duplicate star are in the same order in all tables. The tests pass with two caveats:
- Also, it seems like ~1% of stars with a change in MAG_ACA have virtually no change in MAG_ACA_ERR. I added an ``if`` statement to catch this.
- RSV1 column: it seems that there may be an issue with creating the agasc1p7 h5 file from the fits files which I don't understand. The RSV1 column from fits files does not match the RSV1 column in the original h5 file from the jupyter notebook and ``assert np.all(c_fits == c_jupyter)`` and ``assert np.all(c_fits[ok] == c_1p6[ok])`` both fail on this column.

jeanconn:
> Also, it seems like ~1% of stars with a change in MAG_ACA have virtually no change in MAG_ACA_ERR. I added an if statement to catch this.

I don't understand.  Is this a error I have introduced when I made the fits files, an error in the source h5 file I had from you, or not an error at all?

malgosias:
> > Also, it seems like ~1% of stars with a change in MAG_ACA have virtually no change in MAG_ACA_ERR. I added an if statement to catch this.
> 
> I don't understand. Is this a error I have introduced when I made the fits files, an error in the source h5 file I had from you, or not an error at all?

Cell 87 in
http://nbviewer.jupyter.org/url/cxc.harvard.edu/mta/ASPECT/ipynb/agasc_apass_recalibration.ipynb shows that there will be stars for which the new MAG_ACA_ERR is very close (or the same) as V1.6 MAG_ACA_ERR. I will run this notebook again to confirm that this is the case and count the number of stars for which MAG_ACA_ERR does not change.

malgosias:
> > > Also, it seems like ~1% of stars with a change in MAG_ACA have virtually no change in MAG_ACA_ERR. I added an if statement to catch this.
> > 
> > 
> > I don't understand. Is this a error I have introduced when I made the fits files, an error in the source h5 file I had from you, or not an error at all?
> 
> Cell 87 in
> http://nbviewer.jupyter.org/url/cxc.harvard.edu/mta/ASPECT/ipynb/agasc_apass_recalibration.ipynb shows that there will be stars for which the new MAG_ACA_ERR is very close (or the same) as V1.6 MAG_ACA_ERR. I will run this notebook again to confirm that this is the case and count the number of stars for which MAG_ACA_ERR does not change.

I ran the jupyter notebook and confirmed that there are 170132 stars (in agreement with the number from my test script) for which mag_aca_err change is too small to be reflected in the h5 file where MAG_ACA_ERR column is a small integer. So for example if 1p6 MAG_ACA_ERR is 3 and new MAG_ACA_ERR is 3.3 then the h5 file will still have 3 in this column.

malgosias:
All implemented tests pass.

taldcroft:
Woohoo!

PR #29, opened 2018-09-24
Sort dec

malgosias:
- create_agasc_h5.py - this script creates unsorted h5 file from fits files, with no indexing (agasc1p7_from_fits.h5)
- create_indexed_agasc_h5.py - this script reads in agasc1p7_from_fits.h5 file, sorts it with respect to DEC, and adds RA, DEC, AGASC_ID indexes

### DIFF
--- a/test_agasc1p7.py
+++ b/test_agasc1p7.py
@@ -38,7 +38,7 @@ difflines = """!               THE AXAF GUIDE and ACQUISITION STAR CATALOG V1.6
 ! the AXAF Guide and Acquisition Star Catalog (AGASC) version 1.7
 ! V1.7 changes values for MAG_ACA, MAG_ACA_ERR, RSV1, RSV2 and RSV3""".split('\n')
 
-h5_fits = Table.read('agasc1p7_indexed_from_fits.h5', path='data')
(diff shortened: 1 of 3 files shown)


test_agasc1p7.py
```diff
@@ -38,7 +38,7 @@
 ! the AXAF Guide and Acquisition Star Catalog (AGASC) version 1.7
 ! V1.7 changes values for MAG_ACA, MAG_ACA_ERR, RSV1, RSV2 and RSV3""".split('\n')
 
-h5_fits = Table.read('agasc1p7_from_fits.h5', path='data')
+h5_fits = Table.read('agasc1p7_indexed_from_fits.h5', path='data')
```
jeanconn: File name mismatch here 'agasc1p7_from_fits_indexed.h5' used above.  Looks good to me besides that.

malgosias: Fixed

jeanconn:
And if you remake the miniagasc, it doesn't matter if you use the newly-made sorted or unsorted agasc1p7 file, as long as you use one of the two from this fits process, not the "jupyter" one.

PR #30, opened 2018-09-24
Call out the miniagasc version (1p6) when testing vs DS AGASC

jeanconn:
DS AGASC is set at 1.6, so use miniagasc 1.6 explicitly for the "vs" testing, done via "mp_get_agasc".  This is just a temporary test fix, and might be fine staying as a PR that just shows what changes were needed to get the tests to pass once miniagasc.h5 was set to the 1p7 version.

### DIFF
--- a/agasc/tests/test_agasc_2.py
+++ b/agasc/tests/test_agasc_2.py
@@ -138,7 +138,7 @@ def test_agasc_conesearch(ra, dec):
     _test_agasc(ra, dec)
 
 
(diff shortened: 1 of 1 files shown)


taldcroft:
Use SKA env not proj sot ska.

taldcroft:
@jeanconn - I think this is overtaken by #34, right?

jeanconn:
I also think this is overtaken by #34

PR #33, opened 2018-09-26
Update the docs

jeanconn:
Update the docs for the public-facing fits files

### DIFF
--- a/doc/comments.txt
+++ b/doc/comments.txt
@@ -509,11 +509,7 @@ In AGASC1.7, MAG_ACA is derived from V and BT-VT (COLOR2) and V-i
 (COLOR3) for all stars with valid COLOR2 from Tycho-2 and valid COLOR3
 from AAVSO (American Association of Variable Star Observers) Photometric
 All-Sky Survey (APASS) DR9. MAG_ACA and MAG_ACA_ERR are unchanged from
-AGASC1.6 for stars not meeting these criteria. A linear function was
-fit to the offset between the V-band magnitudes and the observed 
-magnitudes as a function of COLOR2 or COLOR3 for all ACA stars meeting
-the above criteria observed between the start of the mission and day
-2018:150.
+AGASC1.6 for stars not meeting these criteria.
 
 In early 2002, we merged in data from the Tycho-2 catalog, and
 Tycho-2 supplement-1, described further below.  Tycho-2 data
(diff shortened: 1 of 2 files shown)


jeanconn:
I'd like to get these updated docs into the FITS file area and hand off to DS today (though couldn't tell if they actually had the files already via another path from that email yesterday).

doc/comments.txt
```diff
@@ -501,6 +505,16 @@ unchanged from AGASC 1.5 for stars not meeting these criteria.  A
 seven node cubic spline was fit to the offset between the observed
 magnitudes and V-Band magnitudes for 30238 ACA star observations.
 
+In AGASC1.7, MAG_ACA is derived from V and BT-VT (COLOR2) and V-i
+(COLOR3) for all stars with valid COLOR2 from Tycho-2 and valid COLOR3
+from AAVSO (American Association of Variable Star Observers) Photometric
+All-Sky Survey (APASS) DR9.MAG_ACA and MAG_ACA_ERR are unchanged from
+AGASC1.6 for stars not meeting these criteria. A linear function was
```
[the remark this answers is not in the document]
jeanconn: The the fact it is complex suggest to me that we should make it easier to find the docs.  I think there's no link to the agasc1p7 page here and we can do that.

doc/readme.txt
```diff
@@ -350,14 +350,15 @@ BYTES NAME - brief description
...
-NAS8-39073. Information about Chandra and the Chandra X-ray
-Observatory Center may be found on the WWW at http://chandra.harvard.edu/ 
-Detailed information about the catalog and its construction
-can be obtained from the Chandra aspect web page at
-http://asc.harvard.edu/mta/ASPECT/ or by emailing:
+NAS8-39073. Information about Chandra and the Chandra X-ray Observatory
+Center may be found on the WWW at http://chandra.harvard.edu/ 
+Detailed information about the catalog and its construction can be
+obtained from the Chandra aspect web page at
+https://cxc.harvard.edu/mta/ASPECT/agasc1p7 or by emailing:
```
jeanconn: I updated the URLS here to and in the other file to go directly to the agasc1p7 page.  Any objections @malgosias ?

jeanconn: I suppose the agasc link isn't really *the* Chandra aspect web page, but it is the Chandra aspect web page relevant for this product.

PR #36, opened 2018-10-07
Include near neighbor stars and exclude rarely used cols from miniagasc

taldcroft:
For proseco there is legitimate concern about missing stars just above 11.5 mag when considering spoiler stars.  For example a 10.3 candidate acq star would suffer badly from a 11.6 mag star (with a small mag error) that happens to be on the background pixels, and this would currently be accepted as an acq because of more liberal ASPQ1 filtering in proseco.

The main part of this fix is to include in the miniagasc *every* star that is within 60" of a candidate acq or guide star.  That is done via a pre-computed list of such "near neighbor" stars.

Along the way I decided to reduce, not increase, the file size by dropping columns that are basically useless.

### DIFF
--- a/create_near_neighbor_ids.py
+++ b/create_near_neighbor_ids.py
@@ -1,0 +1,51 @@
+# coding: utf-8
+
+"""
+Create a file near_neighbor_ids.fits.gz which contains a single column
+of AGASC IDs corresponding to all stars in AGASC 1.7 that are within
+60 arcsec of a candidate guide or acq star.
+
+This takes a while to run and should be done on a computer with copy
+of AGASC 1.7 on a local (fast) drive.
+
+Usage::
+
+  $ python make_near_neighbor_ids.py
+"""
+import tables
+import os
+from pathlib import Path
+
+import agasc
+from astropy.table import Table
+
+agasc1p7 = str(Path(os.environ['SKA'], 'data', 'agasc', 'agasc1p7.h5'))
+
+h5 = tables.open_file(agasc1p7)
+stars = h5.root.data[:]
+h5.close()
+
+ok = ((stars['CLASS'] == 0) &
+      (stars['MAG_ACA'] < 11.0) &
+      (stars['ASPQ1'] < 50) &  # Less than 2.5 arcsec from nearby star
+      (stars['ASPQ1'] > 0) &
+      (stars['ASPQ2'] == 0)  # Proper motion less than 0.5 arcsec/yr
+      )
+
+# Candidate acq/guide stars with a near neighbor that made ASPQ1 > 0
+nears = stars[ok]
+
+radius = 60 / 3600
+near_ids = set()
+for ii, sp in enumerate(nears):
+    near = agasc.get_agasc_cone(sp['RA'], sp['DEC'], radius=radius, date='2000:001',
+                                agasc_file=agasc1p7)
+    for id in near['AGASC_ID']:
+        if id != sp['AGASC_ID']:
+            near_ids.add(id)
+
+    if ii % 100 == 0:
+        print(ii)
+
+t = Table([list(near_ids)], names=['near_id'])
+t.write('near_neighbor_ids_1p7.fits.gz', format='fits')
--- a/create_mini_agasc_h5.py
+++ b/create_mini_agasc_h5.py
@@ -1,4 +1,5 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+
 import argparse
 import re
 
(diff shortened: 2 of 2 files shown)


taldcroft:
I put the new near neighbors file into /proj/sot/ska/data/agasc for reference.

taldcroft:
And the new file is available at:
/proj/sot/ska/data/agasc/miniagasc_1p7_new.h5

taldcroft:
Not passing tests with proseco but will fix.

jeanconn:
I think the removed columns will be fine with the chandra_aca plot routines (used by starcheck against the miniagasc).  Looks like mica stats and star web app will need a change for the missing columns (either not including those columns or updating  to use the full agasc).

taldcroft:
I had been thinking about plan B, and these other problems you note (thanks!) add to the case for leaving the miniagasc columns as-is but generating a proseco-optimized version that is about half the size of the original.  This makes downloads (for the FOT) faster and improves proseco performance.

taldcroft:
See miniagasc_1p7_new.h5 and proseco_agasc_1p7.h5 in /proj/sot/ska/data/agasc.

jeanconn:
Wasn't sure how fast you were pushing on this.  I think it would be fine to switch mica stats and web page thing over to the full agasc1p7.h5 file (makes sense) so they shouldn't hold back an optimized general miniagasc with fewer columns.  Starcheck plotting could go to a miniagasc without these columns, though I think it would make more sense to swap starcheck over to the full agasc file too.

Regarding the file itself, in general I didn't think it was worth much time to optimize file size for download, as the download is a one* time thing (well, now a two time thing if we want to give them a new file).  How much does it improve performance?

taldcroft:
The FOT are going to get a new file one way or another since putting in the nearby spoiler stars is a high priority (definitely for 4.1).  With current VPN it's taking me 20 minutes to download the full file.  In any case I left the decision on that to Mark B.

I would want to have a discussion about swapping starcheck over to the full AGASC.  If it does not end up using the other columns or any stars for actual checks or plotting, then it winds up just slowing things down.  Also, despite years of feeling otherwise, I've arrived at a desire to run starcheck on my Mac laptop.  With all the work you've done already this seems technically feasible.

jeanconn:
Regarding "I would want to have a discussion about swapping starcheck over to the full AGASC."  Starcheck is using the full agasc for checks and using the miniagasc for plotting.  So it has the worst of both worlds (reading both files).  It should definitely use just one file.  And IIRC the `agasc` module reads the miniagasc and makes the dec index when it starts up even if the user doesn't wan to to use the miniagasc? Whatever.

Anyway, I'd be fine moving starcheck over to a miniagasc, I just haven't had a chance to sufficiently review your new one.  My point was that from an operational perspective, without review, I'd be fine with bringing "swap the plotting to using the full agasc" to LR this week to remove starcheck operational dependence on the default miniagasc.  That would let you do what you like to the miniagasc and we could bring starcheck back over to using it at some later date.  But if you want to keep the proseco version on the side for a bit that's fine too.

create_near_neighbor_ids.py
```diff
@@ -0,0 +1,51 @@
...
+agasc1p7 = str(Path(os.environ['SKA'], 'data', 'agasc', 'agasc1p7.h5'))
+
+h5 = tables.open_file(agasc1p7)
+stars = h5.root.data[:]
+h5.close()
+
+ok = ((stars['CLASS'] == 0) &
+      (stars['MAG_ACA'] < 11.0) &
+      (stars['ASPQ1'] < 50) &  # Less than 2.5 arcsec from nearby star
+      (stars['ASPQ1'] > 0) &
```
jeanconn: Has this been checked?  Meaning is it true that all candidates stars in the AGASC that have a neighbor within 60 arcsecs also have nonzero ASPQ1?  Seems fine but if it is true, we also don't need the filter here, right?  Because removing the filter will add no extra neighbor stars.

create_near_neighbor_ids.py
```diff
@@ -0,0 +1,51 @@
...
+      (stars['ASPQ2'] == 0)  # Proper motion less than 0.5 arcsec/yr
+      )
+
+# Candidate acq/guide stars with a near neighbor that made ASPQ1 > 0
+nears = stars[ok]
+
+radius = 60 / 3600
+near_ids = set()
+for ii, sp in enumerate(nears):
+    near = agasc.get_agasc_cone(sp['RA'], sp['DEC'], radius=radius, date='2000:001',
```
jeanconn: Right, and in general, are all bets off for large PM stars?  We don't generally use them as candidates, but how about as spoilers?  Seems like this "instantaneous" neighbor filter might be a problem for them as well.  Maybe the miniagasc should also just have all stars with proper motion value above some threshold.

jeanconn: Or are all the (recorded in the AGASC) high PM stars bright enough that they are already in the miniagasc by magnitude?

create_near_neighbor_ids.py
taldcroft: What I believe, and the basis for operations since AGASC 1.5, is that no star with ASPQ1 = 0 can have any near neighbor stars that impact tracking or acquisition.  Not sure if we have to reevaluate in light of impact to background pixels, but I think this is unlikely to be a problem in reality.

So skipping ASPQ1 = 0 stars makes the create_near_neighbors script run *way* faster.  Those are the majority of stars.  With this filter in place, it takes about an hour on my laptop, much longer on kadi.  Without, probably 85 hours IIRC.

jeanconn: I just thought the whole point of this exercise was that you were reevaluating in light of the impact of background pixels.

create_near_neighbor_ids.py
taldcroft: Exactly, large PM stars are typically bright because they are by definition nearby.  So in practice I don't believe this will be a problem.  But including all high-PM stars (e.g. > 0.4 arcsec/yr) probably would be a negligible hit on size.

taldcroft:
Good point on starcheck (that's the "having a discussion" part), using the full AGASC with agasc seems fine given your inputs.

jeanconn:
Not quite clear on your last comment, do you want me to get LR approval to use the full agasc for plotting in starcheck this week then?  Or are you fine with having a differently-named miniagasc for proseco testing for now?

taldcroft:
Do what you think makes the most sense given schedule and resources.

taldcroft:
So the actual impact of this PR is to allow creation of the proseco version (which we will install and use) and a near-neighbors version of the miniagasc (which is already created.  Installation is TBD but will require testing to make sure it is a pure superset of the current).

jeanconn:
This looks fine / benign.  Good to merge this too before tagging agasc 4.7 .

taldcroft:
Please hit approve.  :smile:

PR #37, opened 2018-11-08
WIP: Update COLOR1 column to reflect AGASC 1.7 mag updates

taldcroft:
@jeanconn - have a look and see if you agree with this approach.  Trying to find an elegant way to handle the change in AGASC 1.7 content, where mags of some red stars are just fine now, without requiring downstream software changes.


taldcroft:
The problem is that for the stars with "unreliable" mag estimates, the mag error distribution is not at all gaussian.  So I don't see any meaningful way to use MAG_ACA_ERR in that case.  We could put some sentinel value there that means the same thing as COLOR1 = 1.5, but that doesn't help anything.

jeanconn:
I was thinking some penalty that would just make things nicer when I use MAG_ACA_ERR to look for spoiler stars and such in proseco guide star selection.  For star selection, I'm still getting no intuition on if I should use the bright star with huge mag error or the faint star .

taldcroft:
Here's the mag error distribution for red stars without COLOR3 values.  Roughly speaking they are +/- 1.0 mag at 98% confidence, or +1.5, -2.0 mag for "very" high confidence (< 20 out of 3900) were outside that range.  So a 8.0 mag COLOR1=1.5 star (using the updated definition of COLOR1) is definitely better than a 9.8 mag star.  If you use 1-sigma = 1.0 mag (or 1.0 in quadrature with the listed mag_aca_err), that is conservative.

image

agasc/tests/test_agasc_1.py
```diff
@@ -47,3 +48,60 @@ def test_multi_agasc():
     assert len(cone2) == len(cone2_full)
 
 
+def test_update_color1_func():
+    """
+    Test code to update the COLOR1 column.
+    """
+    color1 = [1.0, 1.0, 1.5, 1.5, 1.5, 1.5]
+    color2 = np.array([1.0, 1.0, 1.5, 1.5, 1.75, 2.0]) / 0.850
```
jeanconn: My only nit here would be that color2 is not color2 / 0.850, and condensing this in to one line was a little problem for me with readability.  Not worth a change.

agasc/agasc.py
```diff
@@ -157,6 +186,7 @@ def get_agasc_cone(ra, dec, radius=1.5, date=None, agasc_file=None,
     :param date: Date for proper motion (default=Now)
     :param agasc_file: Mini-agasc HDF5 file sorted by Dec (optional)
     :param pm_filter: Use PM-corrected positions in filtering
+    :param fix_color1: set COLOR1=COLOR2 * 0.85 for stars with V-I color
```
jeanconn: Are all of our COLOR2 values under 2.0 or some such?  I thought there were some limits on the range for which the 0.85 conversion was good, but may not apply to our set.

taldcroft: I don't know the range of COLOR2.  The TYCHO2 readme doesn't indicate any range of applicability, though I didn't read the paper they reference for details.  In any case it doesn't change anything for downstream code or analysis.

Bt_Mag
The Tycho-2 BT magnitude, blank when no magnitude is available. Either BT (bt_mag) or VT (vt_mag) is always given. Approximate Johnson photometric magnitudes may be obtained as follows:

      V   = VT - 0.090*(BT-VT)
      B-V = 0.850*(BT-VT)
  
Consult Sect 1.3 of Vol 1 of "The Hipparcos and Tycho Catalogues", ESA SP-1200, 1997, for more details.

taldcroft:
Tests pass on MacOS.

agasc/agasc.py
jeanconn: Google had given me the Tycho Introduction and Guide to the Data Pg 57

https://www.cosmos.esa.int/documents/532822/552851/vol1_all.pdf/99adf6e3-6893-4824-8fc2-8d3c9cbba2b5

image

jeanconn: Not that it matters much for our uses.

jeanconn:
Will this be 4.7?

PR #40, opened 2019-12-25
Add get_stars() function to get multiple ids / dates

taldcroft:
This is about  25 times faster than running `get_star()` individually when getting large numbers of stars (e.g. more than 1000).   On a fast SSD drive this runs in about 2 msec per star.

Along the way I think I found a way to simplify the code wrt to the proper motion handling.

### DIFF
--- a/agasc/agasc.py
+++ b/agasc/agasc.py
@@ -1,19 +1,18 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import os
+from pathlib import Path
 
 import numpy as np
 import numexpr
 import tables
 
-from ska_path import ska_path
 from Chandra.Time import DateTime
 from astropy.table import Table, Column
 
-__all__ = ['sphere_dist', 'get_agasc_cone', 'get_star']
+__all__ = ['sphere_dist', 'get_agasc_cone', 'get_star', 'get_stars']
 
-DATA_ROOT = ska_path('data', 'agasc')
-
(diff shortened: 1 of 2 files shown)


agasc/agasc.py
```diff
@@ -135,20 +134,15 @@ def add_pmcorr_columns(stars, date):
...
-    # For most of them, the epoch is fixed at 2000, so we don't need N calls to DateTime to
-    # figure that out
-    epoch_is_2000 = (stars['EPOCH'] == 2000.0)
-    ok = has_pm & epoch_is_2000
-    dyear[ok] = (DateTime(date) - DateTime(2000, format='frac_year')) / 365.25
-    # For stars with proper motion correction but epoch != 2000, calculate individually.
-    ok = has_pm & ~epoch_is_2000
-    dyear[ok] = (DateTime(date) - DateTime(stars['EPOCH'][ok], format='frac_year')) / 365.25
-    pm_to_degrees = dyear / (3600. * 1000.)
+    if np.asarray(date).shape == ():
```
jeanconn: @taldcroft Would you please remind me of the expected types and shapes of date?  It isn't clear to me which problem this is code change is solving.

jeanconn: Ah, I see.  The new get_stars() routine repurposes add_pmcorr_columns() to receive date possibly as a thing with the same length as stars.  I think add_pmcorr_columns probably needs a real docstring now and maybe a rename of date to dates in the args.

taldcroft: Documentation update done.  I left the arg as `date` on the off chance this non-public method is being used directly somewhere.

PR #48, opened 2020-10-20
added magnitude estimation (formerly known as mag-stats)

javierggt:
## Description

This is a modified version of what is in the mica mag-stats branch. I renamed most modules and functions.

## Testing

- [ ] Functional testing

Fixes #


agasc/agasc.py
```diff
@@ -353,4 +361,42 @@ def get_stars(ids, agasc_file=None, dates=None, fix_color1=True):
         update_color1_column(t)
     t['DATE'] = dates
 
+    t = _add_supplement(t, supplement_file)
+
     return t if ids.shape else t[0]
+
+
+def _add_supplement(table, supplement_file=None):
```
taldcroft, line 369: For performance:
- The supplement values should be made as a module-level LazyDict. We really need to put this into some common utilities package since it will now be in three places, most recently https://github.com/sot/eng_archive/pull/206.  `ska_helpers`?
- Instead of astropy table join, given the sparse nature of coverage (typically zero to a handful out of hundreds of stars in a field will have data), best to create the columns in advance by copying `MAG_ACA` to `MAG_EST` (and same for the error). Then just loop over the AGASC_ID values and if the id is in the supplement dict then overwrite accordingly.

agasc/agasc.py
```diff
@@ -353,4 +361,42 @@ def get_stars(ids, agasc_file=None, dates=None, fix_color1=True):
...
+    """
+    Add information from AGASC supplement.
+
+    :param table: astropy.table.Table
+    :param supplement_file: str.
```
taldcroft, line 374: Using a `LazyDict` will not allow having this selectable for each query, but for specialized cases the user can always override `DEFAULT_SUPPLEMENT_FILE` in the module before doing the first query.

agasc/agasc.py
```diff
@@ -279,10 +285,12 @@ def get_star(id, agasc_file=None, date=None, fix_color1=True):
     if fix_color1:
         update_color1_column(t)
 
+    t = _add_supplement(t, supplement_file)
+
     return t[0]
 
 
-def get_stars(ids, agasc_file=None, dates=None, fix_color1=True):
+def get_stars(ids, agasc_file=None, dates=None, fix_color1=True, supplement_file=None):
```
taldcroft, line 293: Not needed now.

agasc/agasc.py
```diff
@@ -353,4 +361,42 @@ def get_stars(ids, agasc_file=None, dates=None, fix_color1=True):
         update_color1_column(t)
     t['DATE'] = dates
 
+    t = _add_supplement(t, supplement_file)
```
taldcroft, line 364: Should just be an in-place update `_add_supplement(t)`.

agasc/supplement/magnitudes/star_obs_catalogs.py
```diff
@@ -0,0 +1,242 @@
...
+
+STARS_OBS = None
+"""The table of star observations"""
+
+STARS_OBS_INDICES = None
+STARS_OBS_MAP = None
+STARS_OBS_NP = None
+
+
+def _load_startcat_commands(tstop=None):
```
taldcroft, line 30: Typo in the name?

taldcroft (commented):
Quick first comments, I just realized this is out there and we want to get it into shiny.

taldcroft:
@javierggt - I'm working on a commit to implement my suggestions, so don't worry about this.

taldcroft:
The current dtype of the entries in the `mags` table in the supplement is:
```
      dtype=[('agasc_id', '<i8'), ('color', '<f4'), ('mag_aca', '<f8'), ('mag_aca_err', '<f8'), ('last_obs_time', '<f8')]
```
I would suggest instead:
```
      dtype=[('agasc_id', '<i4'),  ('mag_aca', '<f4'), ('mag_aca_err', '<f4'), ('last_obs_time', '<f8')])
```

taldcroft:
@javierggt - I committed a new approach for doing AGASC queries that also includes tests. I also just realized that having the `use_mag_est` parameter will be helpful in proseco unit testing because we can force that to be False for testing and have reproducible results (always a good thing).

javierggt:
I'm good with that. I will start using this.

taldcroft:
The existing file `/proj/sot/ska/www/ASPECT_ICXC/tmp/jgonzalez/mag_stats/agasc_supplement.h5` does not have the original `bad` table leaf.

agasc/supplement/magnitudes/mag_estimate.py
```diff
@@ -0,0 +1,905 @@
...
+    :param obs_status_override: dict.
+        Dictionary overriding the OK flag for specific observations.
+        Keys are (OBSID, AGASC ID) pairs, values are dictionaries like
+        {'obs_ok': True, 'comments': 'some comment'}
+    :param tstop: cxotime-compatible timestamp
+        Only entries in catalogs.STARS_OBS prior to this timestamp are considered.
+    :return: dict
+        dictionary with stats
+    """
+    min_mag_obs_err = 0.3
```
jeanconn, line 700: Maybe this is the bit that needs to be 0.03?

jeanconn (approved):
With a review of the testing and light review of the agasc/agasc.py changes, this seems fine to me.  I do wonder:
1) would we be better off with downstream fixes instead of the old and new hacks on COLOR1?  
2) do we still want the color1=1.5 acquisition grid model for a star that previously had COLOR1=1.5?  I wasn't sure how much of that selection was based on larger error and how much was based on redder stars (not that we can tease those apart completely)

agasc/supplement/magnitudes/mag_estimate.py
javierggt: Fixed.

taldcroft:
No we do not want the color1=1.5 model for stars that have an estimated mag. The ONLY reason for the distinct model is that the mag_errs for color1=1.5 stars follow a completely different distribution. Once we know the mag of a star then it is no longer in that weird distribution, instead it follows a gaussian distribution (more-or-less) with MAG_ACA_ERR. The red-ness of a star is not relevant once you have a mag.

taldcroft:
> would we be better off with downstream fixes instead of the old and new hacks on COLOR1?

What do you mean by downstream fixes?

jeanconn:
By downstream fixes, I meant changing the handling in proseco sparkles starcheck chandra_aca to be more explicit instead of implicit... basically to remove anything that is using COLOR1=1.5 or COLOR1=0.7 checks.  But I suppose with the current agasc and supplement, we'd be left with a bit of a mess of logic around RSV1, COLOR2, MAG_CATID, unless we added a new column that somewhat defined "does this need extra mag error padding or special handing in the acq model".

I understand that this hack in the agasc code to rewrite COLOR1 works fine with code in the wild like:

https://github.com/sot/proseco/blob/7ad393513387cbe8e08396793ecc5c144f6ace12/proseco/core.py#L1130

just also wondering if that code in proseco really belongs in the agasc module.

Maybe my point in the end is that this agasc change should also kick off an audit/sanity check of how we're using COLOR1 in our code.

javierggt:
I guess downstream would mean changes in proseco.

I added the 'color' column to the supplement, which in principle allows you to specify color in the supplement as well, but I can remove that today since it is not used.

taldcroft:
> I added the 'color' column to the supplement, which in principle allows you to specify color in the supplement as well, but I can remove that today since it is not used.

Yes, color does not belong in the supplement since we do not measure it with the ACA.

javierggt:
removed it.

jeanconn:
Sounds good.  Thanks for having the conversation about it.

agasc/agasc.py
```diff
@@ -177,6 +191,13 @@ def get_agasc_cone(ra, dec, radius=1.5, date=None, agasc_file=None,
     corrected positions are available in the ``RA_PMCORR`` and ``DEC_PMCORR``
     columns, respectively.
 
+    If ``use_mag_est`` is ``True``, then stars with available mag estimates in
+    the AGASC supplement are updated in-place in the output ``stars`` Table:
+
+    - ``MAG_ACA`` and ``MAG_ACA_ERR`` are set according to the supplement.
+    - ``MAG_CATID`` (mag catalog ID) is set to ``MAG_CATID_SUPPLEMENT`` (128).
+    - If COLOR1 is 0.7 or 1.5 then it is changed to 0.69 or 1.49 respectively.
```
jeanconn, line 199: Not sure if we might want a clarifying comment indicating "if the COLOR1 hasn't already been updated based on APASS" or some such.

PR #50, opened 2020-11-12
Nicer documentation including AGASC catalog description

taldcroft:
## Description

Nicer documentation including AGASC catalog description. This removes `readme.txt` as being a redundant subset of the original `comments.txt`.

This PR does not address the issue of documenting the `agasc` package HDF5 files.

## Testing

- [N/A] Passes unit tests on MacOS, linux, Windows (at least one required)
- [x] Functional testing

Docs build cleanly, preview at https://cxc.cfa.harvard.edu/mta/ASPECT/tmp/agasc/


javierggt:
I would have removed the comments.txt and left the readme. The readme is shown in github.

javierggt:
It seems you forgot to add the api.rst file

taldcroft:
The `readme.txt` just applies to the formal FITS AGASC 1.7 catalog, not this package as a whole. I did a diff of the two and `readme.txt` is *mostly* just a subset of `comments.txt`.  So basically I don't want to have two versions of the same thing that need to be kept up to date. We can toss in a `README.md` as needed.

jeanconn:
Looks good.  I was wondering if we might want to add a section to what-was-comments.txt mentioning the h5 version of the files and supplement use by ACA operations and the agasc module to work with it .

taldcroft:
> I was wondering if we might want to add a section to what-was-comments.txt mentioning the h5 version of the files and supplement use by ACA operations and the agasc module to work with it.

In the spirit of smaller PR's that get done in short time, I had explicitly taken that off the table (see the Description). Documenting all the agasc-specific files is a priority, but it will take a little bit of thought and work.

jeanconn:
That's fine.  Probably saw an earlier version of Description and continuing the saga of not knowing when to talk about stuff in the PR vs opening new issue.

javierggt:
Since we are fixing docs, I decided to take the time and add something about the mag-stats. Clearly, it could use some narrative, but I am ok merging just this.

taldcroft:
Looks good. I fixed one section underline that was too short.

PR #51, opened 2020-11-13
Skip the mag supplement tests when no mags in supplement

jeanconn:
## Description

Skip the mag supplement tests when no mags in supplement.

This is somewhat in the style of the other skips in test_agasc_2.  

## Testing

- [x] Passes unit tests on  linux with old supplement (skipping) .  a test with the dev supplement is still failing (not in this change)
- [ ] Functional testing

Fixes #

### DIFF
--- a/agasc/tests/test_agasc_2.py
+++ b/agasc/tests/test_agasc_2.py
@@ -47,6 +47,9 @@ except Exception:
     ascds_env = None
     DS_AGASC_VERSION = None
 
+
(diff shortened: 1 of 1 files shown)


taldcroft (approved):
This looks good, but I'd say to also turn the warning in loading the supplement into a `pass` for now. Which is to say until the supplement is being used operationally. Thanks @jeanconn !

PR #52, opened 2020-11-13
Update supplement MAG_ACA_ERR test

jeanconn:
## Description

Some of the updated values from the supplement matched the catalog
values in the test on the integer MAG_ACA_ERR.  This change updates the test to confirm that the errors are smaller 
or the same coming from the supplement (which may not generalize but appears true here).

I'm still not sure if we have different supplement files in the wild.

## Testing

- [x] Passes unit tests on  linux
- [ ] Functional testing

Fixes #

### DIFF
--- a/agasc/tests/test_agasc_2.py
+++ b/agasc/tests/test_agasc_2.py
@@ -379,7 +379,13 @@ def test_supplement_get_agasc_cone():
     assert np.count_nonzero(stars2['COLOR1'][ok] == 1.49) >= 7
     assert np.count_nonzero(stars2['COLOR1'][ok] == 1.50) == 0
 
-    assert np.all(stars2['MAG_ACA_ERR'][ok] != stars1['MAG_ACA_ERR'][ok])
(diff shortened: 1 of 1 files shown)


javierggt:
I can confirm that with this change the test passes with the new supplement. I'm not sure why test the inequality though. Anyway, better than nothing.

javierggt:
and I can confirm these are the failing tests. It passes when they are skipped.

jeanconn:
Right, I was largely just brainstorming here.  It seems like we are not guaranteed anything about the new errors but they did turn out to be smaller.

jeanconn:
Tom comments: not OK because it allows for all of them to be the same.  So it could be rewritten as np.any(stars2[.] != stars1[..]). I.e. at least some of them are different.  

I'll update with that idea.

jeanconn:
And I just realized I don't know what you meant by stars2[.] and stars1[..] so I just checked the [ok] set.

PR #53, opened 2020-11-13
Remove UserWarning on no mag table in supplement

jeanconn:
## Description

Remove UserWarning on no mag table in supplement

## Testing

- [ ] Functional testing

Fixes #


jeanconn:
I think this one can go without further review.

PR #54, opened 2020-11-17
Fixes in supplement update script (start/stop, input/output, docs)

javierggt:
## Description

This PR introduces a few changes, one of which is a bug fix:
- bug fix: MagStatsReport was renamed MagEstimateReport when moving the project to agasc. Fix one script that was not changed accordingly.
- Added --output-dir, --input-dir, and --reports-dir arguments to the mag-stats scripts.This includes some sensible defaults for weekly runs. Replaced uses of os.path by pathlib.
- Restructured the start/stop arguments so the default matches the weekly run.
- Slightly improved the documentation of the scripts used to update the magnitude supplement.

## Testing

- [x] Functional testing. Ran with several start/stop options (including empty intervals) and output directory options to update supplement and produce reports.
  

Fixes #

### DIFF
--- a/agasc/scripts/mag_estimate_report.py
+++ b/agasc/scripts/mag_estimate_report.py
@@ -24,9 +24,9 @@ def parser():
                                       ' Default: now')
     parse.add_argument('--input-dir', default='$SKA/data/agasc',
                        help='Directory containing mag-stats files. Default: $SKA/data/agasc')
-    parse.add_argument('--output-dir', default='$SKA/www/ASPECT/agasc/supplement_reports/suspect',
+    parse.add_argument('--output-dir', default='supplement_reports/suspect',
                        help='Output directory.'
-                            ' Default: $SKA/www/ASPECT/agasc/supplement_reports/suspect')
+                            ' Default: supplement_reports/suspect')
     parse.add_argument('--obs-stats', default='mag_stats_obsid.fits',
                        help='FITS file with mag-stats for all observations.'
                             ' Default: mag_stats_obsid.fits')
--- a/agasc/scripts/mag_estimate_report.py
+++ b/agasc/scripts/mag_estimate_report.py
@@ -65,9 +65,9 @@ def main():
         directory = args.out_dir
         nav_links = None
 
-    msr = mag_estimate_report.MagStatsReport(agasc_stats,
-                                             obs_stats,
-                                             directory=directory)
+    msr = mag_estimate_report.MagEstimateReport(agasc_stats,
+                                                obs_stats,
+                                                directory=directory)
     msr.multi_star_html(sections=sections, tstart=args.start, tstop=args.stop,
                         filename='index.html',
                         nav_links=nav_links)
--- a/agasc/scripts/mag_estimate_report.py
+++ b/agasc/scripts/mag_estimate_report.py
@@ -22,7 +22,7 @@ def parser():
                                       ' CxoTime-compatible time stamp.')
     parse.add_argument('--input-dir', default='$SKA/data/agasc',
                        help='Directory containing mag-stats files')
-    parse.add_argument('--output-dir', default=f'$SKA/www/agasc_supplement_reports/suspect',
+    parse.add_argument('--output-dir', default=f'$SKA/www/ASPECT/agasc/supplement_reports/suspect',
                        help='Output directory')
     parse.add_argument('--obs-stats', default=f'mag_stats_obsid.fits',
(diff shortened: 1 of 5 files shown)


taldcroft:
> by default, the script updates the supplement in $SKA/data/agasc

In new code I always make the default be "safe" for any output files that get written. I.e. if you just run the script without arguments then nothing can go wrong with the flight files. How you do that depends a bit on the app, but the upshot is that the output "flight" values need to go into `task_schedule.cfg`.  You can look at `acdc` for examples, which mostly have `.` as a default data root directory.

agasc/scripts/mag_estimate_report.py
```diff
@@ -19,7 +20,9 @@ def parser():
                                        ' CxoTime-compatible time stamp.')
     parse.add_argument('--stop', help='Time to stop processing new observations.'
                                       ' CxoTime-compatible time stamp.')
-    parse.add_argument('--out-dir', default=f'./mag_stats_report',
+    parse.add_argument('--input-dir', default='$SKA/data/agasc',
+                       help='Directory containing mag-stats files')
```
taldcroft, line 24: It's generally good to include the default values in the `help` string. (It seems like `argparse` should be able to do that automatically, but I don't recall this functionality).

javierggt:
yes, that makes sense. I will do that instead (default `.`)

agasc/scripts/mag_estimate_report.py
```diff
@@ -19,7 +20,9 @@ def parser():
...
+    parse.add_argument('--output-dir', default=f'$SKA/www/agasc_supplement_reports/suspect',
```
taldcroft, line 25: This path seems weird. I would expect more like `$SKA/www/ASPECT/agasc/agasc_supplement_reports/suspect`, or something like that. 

Also (though this not set in stone), I would usually define a configurable root directory like `$SKA/www/agasc` and then have the rest `agasc_supplement_reports/suspect` hardcoded.  I.e. only define root directories, not detail directories down lower.

agasc/scripts/mag_estimate_report.py
javierggt: Good point. I will do that. For some reason, argparse does not do it. The sphinx-argparse module does, so it shows in web docs, but not in the command-line help.

taldcroft:
@javierggt - it looks like you agreed to the point about not writing into flight dirs by default but the code still does that.

About the output directory, we need to be a little careful about the actual disk that files are being written to. The "originals" should stay on /proj/sot, whereas /proj/sot/ska/www/ASPECT is actually /data/mta4/www/ASPECT (not controlled by us). My first instinct is to write the reports as a sub-directory of /proj/sot/ska/data/agasc so they are more easily discoverable (by me at least) and we can deal with disk space issues if they arise. 

@jeanconn did the same for the dark cals, and demonstrated that a soft link from the /data/mta4 filesystem back to /proj/sot seems to work for the web server (e.g. https://cxc.cfa.harvard.edu/mta/ASPECT/acdc/).

jeanconn:
Regarding being "safe" in where outputs go in new code... I've been somewhat casual in my new code if the code is just reporting/trending and doing nothing that is really operational.  For this code that will have an operational component, I agree with the plan of having it not write into flight areas by default.

javierggt:
I made the change so reports are not saved in /proj/sot/ska/www/ASPECT

agasc/scripts/mag_estimate_report.py
```diff
@@ -7,39 +7,56 @@
...
-    parse.add_argument('--obs-stats', default=f'mag_stats_obsid.fits',
-                       help='FITS file with mag-stats for all observations')
-    parse.add_argument('--agasc-stats', default=f'mag_stats_agasc.fits',
-                       help='FITS file with mag-stats for all observed AGASC stars')
-    parse.add_argument('--weekly-report', help="Add links to navigate weekly reports",
+                                      ' CxoTime-compatible time stamp.'
+                                      ' Default: now')
+    parse.add_argument('--input-dir', default='$SKA/data/agasc',
+                       help='Directory containing mag-stats files. Default: $SKA/data/agasc')
+    parse.add_argument('--output-dir', default='$SKA/www/ASPECT/agasc/supplement_reports/suspect',
```
taldcroft, line 27: Still outputting to a flight directory.

javierggt: oops! It looks like I am actively trying to pass it under the rug...

Fixed it.

taldcroft (approved):

PR #56, opened 2021-01-15
Improved access to AGASC supplement

taldcroft:
## Description

This adds a new function `agasc.get_supplement_table()` which should be the primary external way to access the AGASC supplement. With this it is possible to get any of the 3 tables as either a `Table` or a keyed dict.

It also fixes a bug that showed up in functional testing with `proseco`, namely that the `proseco_agasc_1p7.h5` file doesn't have a `MAG_CATID` so the update code was failing due to a key error.

The function is cached with a one-hour timeout, so a long-lived process will get any updates that occur in the meantime.

## Testing

Testing used the current `agasc_supplement.h5` file from `/proj/sot/ska/jgonzalez/miniconda3/envs/agasc/data/agasc` (if I got that path right from memory), except that I added one entry to the `obs` table.

- [x] Passes unit tests on MacOS
- [x] Functional testing

### Functional testing

Also used this with `proseco` via and sparkles via https://github.com/sot/proseco/pull/345 and `sparkles` via https://github.com/sot/sparkles/pull/149.

### DIFF
--- a/agasc/agasc.py
+++ b/agasc/agasc.py
@@ -17,8 +17,23 @@ __all__ = ['sphere_dist', 'get_agasc_cone', 'get_star', 'get_stars',
            'disable_supplement']
 
 DISABLE_SUPPLEMENT_ENV = 'AGASC_DISABLE_SUPPLEMENT'
-MAG_CATID_SUPPLEMENT = 128
+MAG_CATID_SUPPLEMENT = 100
 RA_DECS_CACHE = {}
+COMMON_DOC = """\
+If ``use_supplement`` is ``True`` and there is no ``AGASC_DISABLE_SUPPLEMENT``
+    environment variable set, then stars with available mag estimates or bad
+    star entries in the AGASC supplement are updated in-place in the output
+    ``stars`` Table:
+
+    - ``MAG_ACA`` and ``MAG_ACA_ERR`` are set according to the supplement.
+    - ``MAG_CATID`` (mag catalog ID) is set to ``MAG_CATID_SUPPLEMENT`` (100).
+    - If ``COLOR1`` is 0.7 or 1.5 then it is changed to 0.69 or 1.49 respectively.
+    - If ``AGASC_ID`` is in the supplement bad stars table then CLASS is set to
+      ``100 + <source>``, where ``<source>`` is the bad star source identifier.
+
+    The default ``agasc_file`` is ``<AGASC_DIR>/miniagasc.h5``, where
+    ``<AGASC_DIR>`` is either the ``AGASC_DIR`` environment variable if defined
+    or ``$SKA/data/agasc``."""
 
 
 def default_agasc_dir():
@@ -109,7 +124,7 @@ def get_supplement_table(name, agasc_dir=None, as_dict=False):
 
 
 def disable_supplement(func):
-    """Decorator to temporarily disable use the AGASC supplement in queries.
+    """Decorator to temporarily disable use of the AGASC supplement in queries.
 
     This is mostly for testing or specialized applications to override the
     default behavior to use the AGASC supplement star mags when available.
@@ -278,17 +293,7 @@ def get_agasc_cone(ra, dec, radius=1.5, date=None, agasc_file=None,
     corrected positions are available in the ``RA_PMCORR`` and ``DEC_PMCORR``
     columns, respectively.
 
-    If ``use_supplement`` is ``True`` and there is no ``AGASC_DISABLE_SUPPLEMENT``
-    environment variable set, then stars with available mag estimates in
-    the AGASC supplement are updated in-place in the output ``stars`` Table:
-
-    - ``MAG_ACA`` and ``MAG_ACA_ERR`` are set according to the supplement.
-    - ``MAG_CATID`` (mag catalog ID) is set to ``MAG_CATID_SUPPLEMENT`` (128).
-    - If COLOR1 is 0.7 or 1.5 then it is changed to 0.69 or 1.49 respectively.
-
-    The default ``agasc_file`` is ``<AGASC_DIR>/miniagasc.h5``, where
-    ``<AGASC_DIR>`` is either the ``AGASC_DIR`` environment variable if defined
-    or ``$SKA/data/agasc``.
+    {common_doc}
 
     Example::
 
@@ -343,24 +348,14 @@ def get_star(id, agasc_file=None, date=None, fix_color1=True, use_supplement=Fal
     """
     Get AGASC catalog entry for star with requested id.
 
-    If ``use_supplement`` is ``True`` and there is no ``AGASC_DISABLE_SUPPLEMENT``
-    environment variable set, then stars with available mag estimates in
-    the AGASC supplement are updated in-place in the output ``stars`` Table:
-
-    - ``MAG_ACA`` and ``MAG_ACA_ERR`` are set according to the supplement.
-    - ``MAG_CATID`` (mag catalog ID) is set to ``MAG_CATID_SUPPLEMENT`` (128).
-    - If COLOR1 is 0.7 or 1.5 then it is changed to 0.69 or 1.49 respectively.
-
-    The default ``agasc_file`` is ``<AGASC_DIR>/miniagasc.h5``, where
-    ``<AGASC_DIR>`` is either the ``AGASC_DIR`` environment variable if defined
-    or ``$SKA/data/agasc``.
+    {common_doc}
 
     Example::
 
       >>> import agasc
       >>> star = agasc.get_star(636629880)
       >>> for name in star.colnames:
-      ...     print '{:12s} : {}'.format(name, star[name])
+      ...     print '{{:12s}} : {{}}'.format(name, star[name])
       AGASC_ID     : 636629880
       RA           : 125.64184
       DEC          : -4.23235
@@ -423,23 +418,13 @@ def get_stars(ids, agasc_file=None, dates=None, fix_color1=True, use_supplement=
     indicating the date at which the star coordinates (RA_PMCORR, DEC_PMCORR)
     are computed.
 
-    If ``use_supplement`` is ``True`` and there is no ``AGASC_DISABLE_SUPPLEMENT``
-    environment variable set, then stars with available mag estimates in
-    the AGASC supplement are updated in-place in the output ``stars`` Table:
-
-    - ``MAG_ACA`` and ``MAG_ACA_ERR`` are set according to the supplement.
-    - ``MAG_CATID`` (mag catalog ID) is set to ``MAG_CATID_SUPPLEMENT`` (128).
-    - If COLOR1 is 0.7 or 1.5 then it is changed to 0.69 or 1.49 respectively.
-
-    The default ``agasc_file`` is ``<AGASC_DIR>/miniagasc.h5``, where
-    ``<AGASC_DIR>`` is either the ``AGASC_DIR`` environment variable if defined
-    or ``$SKA/data/agasc``.
+    {common_doc}
 
     Example::
       >>> import agasc
       >>> star = agasc.get_stars(636629880)
       >>> for name in star.colnames:
-      ...     print '{:12s} : {}'.format(name, star[name])
+      ...     print '{{:12s}} : {{}}'.format(name, star[name])
       AGASC_ID     : 636629880
       RA           : 125.64184
       DEC          : -4.23235
@@ -497,6 +482,11 @@ def get_stars(ids, agasc_file=None, dates=None, fix_color1=True, use_supplement=
         update_from_supplement(t)
 
     return t if ids.shape else t[0]
+
+
+# Interpolate COMMON_DOC into those function docstrings
+for func in get_stars, get_star, get_agasc_cone:
+    func.__doc__ = func.__doc__.format(common_doc=COMMON_DOC)
 
 
 def update_from_supplement(stars):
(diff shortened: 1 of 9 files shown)


jeanconn:
I know I'm the most pedantic of us, but do we need/want a way to identify which version of the supplement was used?  And would that be just a date key or something?

taldcroft:
I guess the issue is that the AGASC supplement is not versioned and it is not possible to roll back to an arbitrary version even if you have a version / date. So right now we basically have "use the latest/greatest supplement", or "don't use the supplement at all for reproducibility". 

I hear you that this might be a nice-to-have, but it doesn't quite fit into the current data model we've established. Unless there is something I'm not considering.

taldcroft:
It occurred to me in the meantime that we could probably just put the whole file into a git repo and make a commit and tag at the point of each "release" (promotion). The version can be stored in the file in a HDF5 keyword. 

I'd be happy enough with that and the `use_supplement` keyword could even be allowed to take a string indicating the version (tag) to use.

javierggt:
I think it is good, in principle, to be able to somehow keep track of the supplement version. I would not recommend adding the supplement itself to version control unless we change the format.

Currently, the file is about 1.5 GB, and we intend to modify it once a week. If git keeps a copy of each version, that is 75 GB in a year. And this gets downloaded when you clone the repository. If we still want to keep the format and add it to git, one option could be to use `git lfs`. I'm not sure that is a good idea though.

taldcroft:
I'm showing the `agasc_supplement.h5` file as 5.9 Mb, and it will probably be close to half that size when you change the dtypes. I was only thinking of putting that one file in version control.

taldcroft:
Also, that repo would not be part of the `agasc` project repo on GH, just a local git repo that allows us to look back in time at versions.

jeanconn:
Right, I don't have a huge justification for the additional work/tracking version, but I was disturbed at the thought of not being able to easily reproduce star selection.  I had been thinking of either a "save last week's file" method or just tracking of date so one could regenerate a supplement as it would have existed at a date.  The git repo of the file with tags for weekly releases sounds better.  I note that I can't think of a reason to distribute the repo  (unlike, say, the new thermal models version control with git repo as the sync'd data product).  Which I mention not because it was suggested... just talking this out in the comments.

javierggt:
ok, so it seems we will use a separate repository for versioning the supplement. If so, then we can use the commit SHA or the version returned by ska_helpers, but then the update scripts should include committing to the repository.

agasc/agasc.py
```diff
@@ -1,42 +1,178 @@
...
+If ``use_supplement`` is ``True`` and there is no ``AGASC_DISABLE_SUPPLEMENT``
+    environment variable set, then stars with available mag estimates or bad
+    star entries in the AGASC supplement are updated in-place in the output
+    ``stars`` Table:
+
+    - ``MAG_ACA`` and ``MAG_ACA_ERR`` are set according to the supplement.
+    - ``MAG_CATID`` (mag catalog ID) is set to ``MAG_CATID_SUPPLEMENT`` (100).
+    - If ``COLOR1`` is 0.7 or 1.5 then it is changed to 0.69 or 1.49 respectively.
+    - If ``AGASC_ID`` is in the supplement bad stars table then CLASS is set to
+      ``100 + <source>``, where ``<source>`` is the bad star source identifier.
```
javierggt, line 34: I don't like having hardwired numbers in the code. You have added three. I can understand the replacements for 0.7 and 1.5 because the idea is that now they would not be the old hardwired numbers and you can still use COLOR1 as a proxy for large uncertainty. Not pretty but ok.

I do not understand why add the hardwired source. Couldn't you have a "bad" column?  This kind convention (100 + source) is not good style, so why do it in the first place if it is not necessary? It is not needed for backward compatibility, is it? and having a "bad" column would be explicit.

agasc/agasc.py
```diff
@@ -1,42 +1,178 @@
...
+    """Default main AGASC file ``agasc_dir() / miniagasc.h5``.
+
+    :returns: str
+    """
+    return str(default_agasc_dir() / 'miniagasc.h5')
 
 
-__all__ = ['sphere_dist', 'get_agasc_cone', 'get_star', 'get_stars', 'MAG_CATID_SUPPLEMENT']
+@lru_cache_timed(timeout=3600)
+def get_supplement_table(name, agasc_dir=None, as_dict=False):
```
javierggt, line 65: Originally, I thought this function would be in the supplement namespace. I.e. `agasc.supplement.get_table` instead of `agasc.get_supplement_table`

agasc/agasc.py
```diff
@@ -397,29 +509,67 @@ def get_stars(ids, agasc_file=None, dates=None, fix_color1=True, use_mag_est=Fal
...
 
 
-def update_mags_from_supplement(stars):
-    """Overwrite mag and color1 information from AGASC supplement in ``stars``.
+# Interpolate COMMON_DOC into those function docstrings
+for func in get_stars, get_star, get_agasc_cone:
+    func.__doc__ = func.__doc__.format(common_doc=COMMON_DOC)
+
+
+def update_from_supplement(stars):
```
javierggt, line 523: And this is another that could be in the supplement namespace. I.e. `agasc.supplement.update_star` instead of `agasc.update_from_supplement`

agasc/agasc.py
```diff
@@ -397,29 +509,67 @@ def get_stars(ids, agasc_file=None, dates=None, fix_color1=True, use_mag_est=Fal
...
-            stars['MAG_CATID'][idx] = MAG_CATID_SUPPLEMENT
-            color1 = stars['COLOR1'][idx]
-            if np.isclose(color1, 0.7) or np.isclose(color1, 1.5):
-                stars['COLOR1'][idx] = color1 - 0.01
+            set_star(star, 'MAG_ACA_ERR', round(mag_est_err * 100))
+            set_star(star, 'MAG_CATID', MAG_CATID_SUPPLEMENT)
+            if 'COLOR1' in stars.colnames:
+                color1 = star['COLOR1']
+                if np.isclose(color1, 0.7) or np.isclose(color1, 1.5):
+                    star['COLOR1'] = color1 - 0.01
```
jeanconn, line 572: This seems fine overall, though I note it may be a little confusing for / will not apply to stars that have already been "fixed up" by the update_color1_column method.

agasc/agasc.py
```diff
@@ -1,42 +1,178 @@
...
+    star entries in the AGASC supplement are updated in-place in the output
+    ``stars`` Table:
+
+    - ``MAG_ACA`` and ``MAG_ACA_ERR`` are set according to the supplement.
+    - ``MAG_CATID`` (mag catalog ID) is set to ``MAG_CATID_SUPPLEMENT`` (100).
+    - If ``COLOR1`` is 0.7 or 1.5 then it is changed to 0.69 or 1.49 respectively.
+    - If ``AGASC_ID`` is in the supplement bad stars table then CLASS is set to
+      ``100 + <source>``, where ``<source>`` is the bad star source identifier.
+
+    The default ``agasc_file`` is ``<AGASC_DIR>/miniagasc.h5``, where
```
jeanconn, line 36: Seems like the common doc section should also mention the filename of the agasc supplement.  I agree that we can probably get by without adding an option to control / specify that by name.

agasc/agasc.py
taldcroft: I understand your concerns. This code falls in the category of "perfect is the enemy of good".

Using `CLASS` instead of a new column is needed for backward compatibility. Proseco, starcheck, and MATLAB tools (SAUSAGE) use `CLASS` as a way of identifying whether a star might be acceptable. For instance `CLASS` is in the star information hover-over now in starcheck and impacts how stars are plotted. So a CLASS of 100 or more will ensure that existing code will recognize that star as (generally speaking) not accepted for star selection. 

I am perfectly happen to just use `BAD_CLASS_SUPPLEMENT` always, so unless @jeanconn objects I'll do that.

Another driver here is that I might like to clean up `proseco` so it doesn't specifically look at the `bad` HDF5 table, but instead just defers to `agasc` to tell that a star is bad based on CLASS. This probably already happens for free with no code change.

agasc/agasc.py
taldcroft: Will do.

agasc/agasc.py
taldcroft: I can do that, maybe like this:
- Move `get_supplement_table` into its own module in `agasc/supplement/utils.py` (better name?). There is no need to load all the other modules that all relate to maintaining the supplement rather than user access.
- Factor out the path definitions so that `default_agasc_dir` is available to `agasc/supplement/utils.py` without a circular import. Putting paths into a separate `paths.py` module could be a good thing anyway if it gets used elsewhere in the package.
- Update the package `__init__.py` to import the function from `agasc.supplement.utils`.

agasc/agasc.py
taldcroft: This is really an internal helper function for `get_stars` and `get_cone_agasc`. It is tightly integrated with those functions and not exported from the module. So I'd rather keep it here in the module that is providing user access to the AGASC.

agasc/agasc.py
javierggt: sounds good to me.

agasc/agasc.py
taldcroft: Good catch on a possible interaction that I didn't notice, but it looks like this is fine. The color update runs first and might change the 1.5 to something else, in which case this code won't do anything, which is fine. Basically we just need to un-1.5-ify that color one way or another. If only the AGASC had come with another status column 25 years ago we wouldn't be playing these games.

jeanconn: Right, I mean, we talked a bit about updating starcheck and proseco to just not *use* color1  but I still don't have any great alternative ideas at this point anyway.

taldcroft: The real truth is that COLOR1 is already effectively a status column. In other words, we don't actually use the value of color anywhere in operations or analysis apart from checking for 0.7 and 1.5, and of course estimating MAG_ACA in the AGASC catalog generation. From the operations perspective it is basically a 3-valued status flag, "OK", "Unreliable mag - no color", and "Unreliable mag - red". 

We could formalize this by defining exported constants in `agasc`, if that would help things, but the name of this "status" column will always be misleading.

jeanconn: Good points.  I was thinking more that for me the confusion is that the 'OK' status can be either 'OK color in original catalog', 'OK color due to update from AGASC 1.7 color information', 'OK color from due to mag estimates previous observations'.  Are all of the MAG_ACA_ERR estimates equally good for those 3 "OK" cases?  

As a tiny nit, I suppose this brings up the issue that I had pushed to get the mag supplement to have an entry for all color1=1.5 stars if observed.  I think if the mag supplement code is identifying color1=1.5 stars based on the agasc get_star method with fix_color1=True, we don't automatically get new mag estimates for stars that were updated via the AGASC 1.7 color updates.  Do we care?  Maybe we don't.  We get new estimates for all stars if "different enough" from catalog and the "color1 updated to be not 1.5 due to AGASC 1.7" stars already had their color1 penalties removed in star selection.

taldcroft: > Are all of the MAG_ACA_ERR estimates equally good for those 3 "OK" cases?

Yes. Note that I don't see any reason to ever use `fix_color1=False` in operations or analysis code. It's there for testing / development to see what's really in the AGASC.

> I had pushed to get the mag supplement to have an entry for all color1=1.5 stars if observed

The mag supplement now has an entry for *every* star if it has been observed, and we are not applying any "different enough" criteria for inclusion. That was an early idea but it got dropped IIRC. At least this is all my understanding, given that the supplement has 81000 stars in it and most of them are very close to the original.

jeanconn: Ah.  OK.  I guess I got a bit confused about that wrt the highlighting in the report vs the actual supplement content at this point.

agasc/agasc.py
taldcroft: Fixed, to the extent practical.

agasc/agasc.py
taldcroft: Done.

taldcroft:
I think I have addressed all the comments now, and also made the main documentation page a bit better.

agasc/agasc.py
```diff
@@ -182,7 +240,7 @@ def add_pmcorr_columns(stars, date):
 
 
 def get_agasc_cone(ra, dec, radius=1.5, date=None, agasc_file=None,
-                   pm_filter=True, fix_color1=True, use_mag_est=False):
+                   pm_filter=True, fix_color1=True, use_supplement=False):
```
jeanconn, line 243: As we dig more into implementation... what would the disadvantages be at this point to flipping the default?

taldcroft: Basically any existing code that uses AGASC mags will no longer give reproducible results. I'm a bit wary of that.

jeanconn: Right, I completely agree it is a scary idea, but I'm not convinced there is code out there for which it matters.

taldcroft: OK, let's not be so timid! I guess there is precedent with fixing star colors by default. I have also globally enabled this by default.

agasc/agasc.py
```diff
@@ -1,42 +1,93 @@
...
+    Examples::
+
+      import agasc
+
+      # Disable use of the supplement
+      with agasc.set_supplement_enabled(False):
+          aca = proseco.get_aca_catalog(obsid=8008)
+
+      # Allow using the supplement
+      @agasc.set_supplement_enabled(False)
```
jeanconn, line 69: Little confused about this example.

taldcroft: Oops, fixed.

PR #58, opened 2021-01-24
Faster get_stars() function

taldcroft:
## Description

For getting stars by AGASC_ID, at some point it is faster to just read the entire AGASC catalog into memory. This PR implements that method using a threshold of 5000 stars. That value was determined based on testing on my laptop.

## Testing

- [x] Passes unit tests on MacOS.
- [x] Functional testing

### Functional testing

For the AGASC supplement with about 81000 stars, the original (`tables.read_where()`) method takes about 80 seconds, while the new method takes about 13 seconds.

### DIFF
--- a/agasc/agasc.py
+++ b/agasc/agasc.py
@@ -352,13 +352,54 @@ def get_star(id, agasc_file=None, date=None, fix_color1=True, use_supplement=Non
     return t[0]
 
 
-def get_stars(ids, agasc_file=None, dates=None, fix_color1=True, use_supplement=None):
+def _get_rows_read_where(ids_1d, dates_1d, agasc_file):
+    rows = []
+    with tables.open_file(agasc_file) as h5:
+        tbl = h5.root.data
+        for id, date in zip(ids_1d, dates_1d):
+            id_rows = tbl.read_where('(AGASC_ID == {})'.format(id))
+
+            if len(id_rows) > 1:
+                raise InconsistentCatalogError(
+                    f'More than one entry found for {id} in AGASC')
+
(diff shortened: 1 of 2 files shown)


agasc/agasc.py
```diff
@@ -366,33 +406,32 @@ def get_stars(ids, agasc_file=None, dates=None, fix_color1=True, use_mag_est=Fal
     :param dates: Dates for proper motion (scalar or array) (default=Now)
     :param fix_color1: set COLOR1=COLOR2 * 0.85 for stars with V-I color (default=True)
     :param use_mag_est: Use estimated mag from AGASC supplement where available (default=False)
+    :param read_entire_agasc: Number of stars above which to read the entire AGASC (default=5000)
```
jeanconn, line 409: Love that you set the number of stars as function parameter, though the name sounds more like it should be a flag/boolean.

jeanconn (approved):
This seems fine to me.  I noted that the option might not have the best name but I don't consider that blocking.

agasc/agasc.py
javierggt: I kind of agree with Jean that the name suggests something else to me.

I would call it `cache_threshold` or something like that.

taldcroft: Changed to `method_threshold`.  

Note - there is no caching involved, on purpose. I don't want to leave the whole AGASC and index dict in memory by default. Maybe a subsequent PR could allow that if it turns out to be useful. In my work the large `get_stars` calls are just one-offs.

taldcroft:
Note that the flake8 issues are unrelated to this PR.

PR #85, opened 2021-02-11
Added supplement versions table

### LINKED ISSUE #74: Add a version to the supplement
Add a version to the supplement 

Tom suggests adding the date/day of the processing/update to the metadata.

javierggt:
## Description

This PR adds two tables to the supplement:
- versions
- last_updated

They have the same structure. Each is a one-row table with string dtype. It is easy to remove the `last_updated` table if you prefer not to have it.

I thought of the following options:
- saving as meta data for each table. Jean wanted a global version as well, plus there should be a method to retrieve versions, and we already have `utils.get_supplement_table`
- saving in a single table, with one row for each existing table, one column for version and another for last_updated. This is fine. It just meant the user code had to jump through some hoops to get the version.

## Testing

- [x] Functional testing. You can run this:
   ```
   (ska3-flight-2021.2rc4) javierg agasc $ export AGASC_DIR=`pwd`/test
   (ska3-flight-2021.2rc4) javierg agasc $ ipython
   Python 3.8.3 (default, Jul  2 2020, 11:26:31) 
   Type 'copyright', 'credits' or 'license' for more information
   IPython 7.16.1 -- An enhanced Interactive Python. Type '?' for help.
   histogram not found

   In [1]: import logging 
      ...: logging.basicConfig(level='DEBUG') 
      ...: import agasc 
      ...: from agasc.supplement.utils import save_version, get_supplement_table 
      ...: from agasc.supplement.utils import save_version, get_supplement_table 
      ...: filename = 'test/agasc_supplement.h5' 
      ...: save_version(filename, obs=agasc.__version__, mags=agasc.__version__, bad=agasc.__version__) 
      ...:                                                                                                                                                      
   DEBUG:agasc.supplement:Creating agasc supplement table "versions"
   DEBUG:agasc.supplement:Creating agasc supplement table "last_updated"
   DEBUG:agasc.supplement:Adding "obs" to agasc supplement "versions" table
   DEBUG:agasc.supplement:Adding "obs" to agasc supplement "last_updated" table
   DEBUG:agasc.supplement:Adding "mags" to agasc supplement "versions" table
   DEBUG:agasc.supplement:Adding "mags" to agasc supplement "last_updated" table
   DEBUG:agasc.supplement:Adding "bad" to agasc supplement "versions" table
   DEBUG:agasc.supplement:Adding "bad" to agasc supplement "last_updated" table
   DEBUG:agasc.supplement:Adding "supplement" to agasc supplement "versions" table
   DEBUG:agasc.supplement:Adding "supplement" to agasc supplement "last_updated" table
   In [2]: get_supplement_table('versions')                                                                                                                     
   Out[2]: 
   <Table length=1>
            obs                   mags                  bad                supplement     
          bytes32               bytes32               bytes32               bytes32       
   --------------------- --------------------- --------------------- ---------------------
   4.10.3.dev89+g8c21663 4.10.3.dev89+g8c21663 4.10.3.dev89+g8c21663 4.10.3.dev89+g8c21663

   In [3]: get_supplement_table('last_updated')                                                                                                                 
   Out[3]: 
   <Table length=1>
             obs                     mags                    bad                  supplement      
           bytes32                 bytes32                 bytes32                 bytes32        
   ----------------------- ----------------------- ----------------------- -----------------------
   2021-02-11 19:11:48.412 2021-02-11 19:11:48.412 2021-02-11 19:11:48.412 2021-02-11 19:11:48.412
   ```

Fixes #74

### DIFF
--- a/agasc/scripts/update_obs_status.py
+++ b/agasc/scripts/update_obs_status.py
@@ -24,7 +24,7 @@ from agasc.supplement.utils import save_version
 logger = logging.getLogger('agasc.supplement')
 
 
-def add_bad_star(bad_star_ids, bad_star_source, suppl_file, dry_run):
+def add_bad_star(bad_star_ids, bad_star_source, suppl_file, dry_run, create=False):
     """
     Update the 'bad' table of the AGASC supplement.
 
@@ -32,14 +32,19 @@ def add_bad_star(bad_star_ids, bad_star_source, suppl_file, dry_run):
     :param bad_star_source: int or list
     :param suppl_file: pathlib.Path
     :param dry_run: bool
+    :param create: bool
+        Create a supplement file if it does not exist
     """
+    suppl_file = Path(suppl_file)
+
     if not bad_star_ids:
         logger.info('Nothing to update')
         return
 
-    if not Path(suppl_file).exists():
+    if not suppl_file.exists():
+        if not create:
+            raise FileExistsError(suppl_file)
         logger.warning(f'Creating a new AGASC supplement: {suppl_file}')
-        raise FileExistsError(suppl_file)
 
     logger.info(f'updating "bad" table in {suppl_file}')
 
@@ -77,11 +82,11 @@ def add_bad_star(bad_star_ids, bad_star_source, suppl_file, dry_run):
     logger.info('The wiki page also includes instructions for test, review, approval')
     logger.info('and installation.')
     if not dry_run:
-        dat.write(str(suppl_file), format='hdf5', path='bad', append=True, overwrite=True)
-        save_version(suppl_file, obs=agasc.__version__)
+        dat.write(suppl_file, format='hdf5', path='bad', append=True, overwrite=True)
+        save_version(suppl_file, bad=agasc.__version__)
 
 
-def update_obs_table(filename, obs_status_override, dry_run=False):
+def update_obs_table(filename, obs_status_override, dry_run=False, create=False):
     """
     Update the 'obs' table of the AGASC supplement.
 
@@ -91,14 +96,17 @@ def update_obs_table(filename, obs_status_override, dry_run=False):
         Keys are (OBSID, AGASC ID) pairs, values are dictionaries like
         {'status': 0, 'comments': 'some comment'}
     :param dry_run: bool
+    :param create: bool
+        Create a supplement file if it does not exist
     """
     if not obs_status_override:
         logger.info('Nothing to update')
         return
 
     if not Path(filename).exists():
+        if not create:
+            raise FileExistsError(filename)
         logger.warning(f'Creating a new AGASC supplement: {filename}')
-        raise FileExistsError(filename)
 
     try:
         obs_status = table.Table.read(filename, format='hdf5', path='obs')
@@ -140,7 +148,7 @@ def update_obs_table(filename, obs_status_override, dry_run=False):
         obs_status = table.Table(dtype=obs_dtype)
 
     if not dry_run:
-        obs_status.write(str(filename), format='hdf5', path='obs', append=True, overwrite=True)
+        obs_status.write(filename, format='hdf5', path='obs', append=True, overwrite=True)
         save_version(filename, obs=agasc.__version__)
     else:
         logger.info('dry run, not saving anything')
--- a/agasc/scripts/update_obs_status.py
+++ b/agasc/scripts/update_obs_status.py
@@ -81,7 +81,7 @@ def add_bad_star(bad_star_ids, bad_star_source, suppl_file, dry_run, create=Fals
     logger.info('The wiki page also includes instructions for test, review, approval')
     logger.info('and installation.')
     if not dry_run:
-        dat.write(suppl_file, format='hdf5', path='bad', append=True, overwrite=True)
+        dat.write(str(suppl_file), format='hdf5', path='bad', append=True, overwrite=True)
         save_version(suppl_file, 'bad')
 
 
--- a/agasc/scripts/update_obs_status.py
+++ b/agasc/scripts/update_obs_status.py
@@ -83,7 +83,7 @@ def add_bad_star(bad_star_ids, bad_star_source, suppl_file, dry_run, create=Fals
     logger.info('and installation.')
     if not dry_run:
         dat.write(suppl_file, format='hdf5', path='bad', append=True, overwrite=True)
-        save_version(suppl_file, bad=agasc.__version__)
+        save_version(suppl_file, 'bad')
 
 
 def update_obs_table(filename, obs_status_override, dry_run=False, create=False):
@@ -149,7 +149,7 @@ def update_obs_table(filename, obs_status_override, dry_run=False, create=False)
 
     if not dry_run:
         obs_status.write(filename, format='hdf5', path='obs', append=True, overwrite=True)
-        save_version(filename, obs=agasc.__version__)
+        save_version(filename, 'obs')
     else:
         logger.info('dry run, not saving anything')
 
--- a/agasc/scripts/update_obs_status.py
+++ b/agasc/scripts/update_obs_status.py
@@ -17,7 +17,9 @@ import numpy as np
 from astropy import table
 import pyyaks.logger
 
+import agasc
 from agasc.supplement.magnitudes import star_obs_catalogs
+from agasc.supplement.utils import save_version
 
 logger = logging.getLogger('agasc.supplement')
 
@@ -76,6 +78,7 @@ def add_bad_star(bad_star_ids, bad_star_source, suppl_file, dry_run):
     logger.info('and installation.')
     if not dry_run:
         dat.write(str(suppl_file), format='hdf5', path='bad', append=True, overwrite=True)
+        save_version(suppl_file, obs=agasc.__version__)
 
 
 def update_obs_table(filename, obs_status_override, dry_run=False):
@@ -138,6 +141,7 @@ def update_obs_table(filename, obs_status_override, dry_run=False):
 
     if not dry_run:
         obs_status.write(str(filename), format='hdf5', path='obs', append=True, overwrite=True)
+        save_version(filename, obs=agasc.__version__)
     else:
         logger.info('dry run, not saving anything')
 
--- a/agasc/scripts/update_obs_status.py
+++ b/agasc/scripts/update_obs_status.py
@@ -17,7 +17,6 @@ import numpy as np
 from astropy import table
 import pyyaks.logger
 
-import agasc
 from agasc.supplement.magnitudes import star_obs_catalogs
 from agasc.supplement.utils import save_version
 
--- a/agasc/supplement/utils.py
+++ b/agasc/supplement/utils.py
@@ -1,15 +1,20 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 from pathlib import Path
+import logging
 import warnings
 
 from ska_helpers.utils import lru_cache_timed
 import tables
-from astropy.table import Table
+from cxotime import CxoTime
+from astropy.table import Table, Column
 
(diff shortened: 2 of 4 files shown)


agasc/supplement/utils.py
```diff
@@ -73,3 +79,60 @@ def get_supplement_table(name, agasc_dir=None, as_dict=False):
...
+        from agasc.supplement.utils import save_version
+        save_version('agasc_supplement.h5', mags='4.10.2')
+
+    The different versions can be retrieved from the default supplement::
+
+        from agasc.supplement.utils import get_supplement_table
+        versions = get_supplement_table('versions')
+
+    :param filename: pathlib.Path
+    :param kwargs: dict
```
taldcroft, line 112: Since this function is always being called (except in testing) with `save_version(filename, <table_name>=agasc.__version__)`, maybe just simplify to `save_version(filename, table_name)`?

agasc/supplement/utils.py
```diff
@@ -73,3 +79,60 @@ def get_supplement_table(name, agasc_dir=None, as_dict=False):
...
+    version_dtype = '<U32'
+    filename = Path(filename)
+
+    versions = _load_or_create(filename, 'versions')
+    last_updated = _load_or_create(filename, 'last_updated')
+
+    time = CxoTime().iso
+    kwargs.update({'supplement': agasc.__version__})
+    for key, value in kwargs.items():
+        pass
```
taldcroft, line 125: Leftover.

agasc/supplement/utils.py
```diff
@@ -73,3 +79,60 @@ def get_supplement_table(name, agasc_dir=None, as_dict=False):
...
+
+    :param filename: pathlib.Path
+    :param kwargs: dict
+        A dictionary of table name/version pairs,
+    """
+    import agasc
+    version_dtype = '<U32'
+    filename = Path(filename)
+
+    versions = _load_or_create(filename, 'versions')
```
taldcroft, line 119: Can we name this table `agasc_versions`? I got quite confused because I think of `versions` as being versions of the tables, not versions of the package that wrote the tables.

agasc/supplement/utils.py
```diff
@@ -73,3 +79,60 @@ def get_supplement_table(name, agasc_dir=None, as_dict=False):
...
+    last_updated = _load_or_create(filename, 'last_updated')
+
+    time = CxoTime().iso
```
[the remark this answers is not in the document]
taldcroft: Need to check, but I think you can reduce much of the code in the function code to basically:
```
logger.debug(f'Updating {key} in "versions" and "last_updated" tables')
for key, value in kwargs.items():
   versions[key] = [value]
   last_updated[key] = [time]
```
You don't need the `'unknown'`, nor need to pre-declare as columns with a dtype. The key point is that astropy will add in the column or else replace it as the case may be. And I think that the HDF5 writing step will just overwrite whatever is there.

agasc/supplement/utils.py
```diff
@@ -43,8 +48,9 @@ def get_supplement_table(name, agasc_dir=None, as_dict=False):
     """
     agasc_dir = default_agasc_dir() if agasc_dir is None else Path(agasc_dir)
 
-    if name not in ('mags', 'bad', 'obs'):
-        raise ValueError("table name must be one of 'mags', 'bad', or 'obs'")
+    if name not in ('mags', 'bad', 'obs', 'last_updated', 'versions'):
```
taldcroft, line 51: At this point it may be worth defining a module `AGASC_SUPPLEMENT_TABLES` constant with the names of all the available tables.

agasc/tests/test_obs_status.py
```diff
@@ -434,3 +436,26 @@ def recreate_test_supplement():
     update_obs_status.update_obs_table(TEST_DATA_DIR / 'agasc_supplement.h5',
                                        obs_status_override,
                                        dry_run=False)
+
+
+def test_save_version(monkeypatch):
```
taldcroft, line 441: Nice with the monkey patching!  I haven't really gone down this path before but it looks good.

agasc/scripts/update_obs_status.py
```diff
@@ -76,6 +78,7 @@ def add_bad_star(bad_star_ids, bad_star_source, suppl_file, dry_run):
     logger.info('and installation.')
     if not dry_run:
         dat.write(str(suppl_file), format='hdf5', path='bad', append=True, overwrite=True)
+        save_version(suppl_file, obs=agasc.__version__)
```
taldcroft, line 81: I think this should be `bad` not `obs`.

agasc/supplement/utils.py
```diff
@@ -73,3 +79,60 @@ def get_supplement_table(name, agasc_dir=None, as_dict=False):
...
+        if key not in versions.colnames:
+            logger.debug(f'Adding "{key}" to agasc supplement "versions" table')
+            versions[key] = Column(data=['unknown'], dtype=version_dtype)
+        if key not in last_updated.colnames:
+            logger.debug(f'Adding "{key}" to agasc supplement "last_updated" table')
+            last_updated[key] = Column(data=['unknown'], dtype=version_dtype)
+        versions[key][:] = value
+        last_updated[key][:] = time
+
+    versions.write(str(filename), format='hdf5', path='versions',
```
taldcroft, line 135: Astropy will accept a `Path`, no need for `str()`here.

taldcroft (commented):
One actual bug (I think), an API suggestion and ideas for simplifying / reducing the code (less is more!). Mostly looks good though.

agasc/supplement/utils.py
```diff
@@ -73,3 +79,60 @@ def get_supplement_table(name, agasc_dir=None, as_dict=False):
...
+
+def _load_or_create(filename, table_name):
+    try:
+        table = Table.read(filename, format='hdf5', path=table_name)
+    except OSError as e:
+        logger.debug(f'Creating agasc supplement table "{table_name}"')
+        if not filename.exists() or str(e) == f'Path {table_name} does not exist':
+            table = Table()
+        else:
+            raise
```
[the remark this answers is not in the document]
javierggt: Actually, the code you propose is not equivalent. What you propose checks if the file exists, while the current code checks whether the table exists in the file.

I agree that relying on exception messages is fragile, but Table.read does not raise a specific exception. One way could be to not use astropy.Table and instead use h5py or pytables to check for the existence of the table.

agasc/supplement/utils.py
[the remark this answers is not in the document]
javierggt: Changed to use `CxoTime.now` and set precision.

> You don't need ... to pre-declare as columns with a dtype.

well, I have gotten upset by astropy changing the types when saving/loading fits files (which might be caused by fits, but still) so I decided to have all table types explicit. Anyway, I changed it according to the suggestion.

agasc/supplement/utils.py
javierggt: yes, I thought of that, and I also though of having a default argument. I can remove the argument because it makes it simpler, and one can always add it if you want to use a custom version string.

Are you sure you will not want a custom version string? I changed it according to the suggestion, but one can still undo that commit.

agasc/supplement/utils.py
javierggt: Fixed

agasc/supplement/utils.py
javierggt: I changed the table name. We could also add the table version itself, but this is another case where scope is creeping up. The next obvious question is whether we want to track table versions and guarantee back compatibility with older table versions. If we wanted to do that, then perhaps the supplement would have been implemented differently from the ground up.

agasc/supplement/utils.py
javierggt: Done

agasc/tests/test_obs_status.py
javierggt: I think we should do more of that in unit testing so it does not depend on the actual SKA/data, and leave the use of SKA/data for integration testing. I think one can also use monkeypatch to generate the data files we would use (say, for cheta)

agasc/scripts/update_obs_status.py
javierggt: Fixed

agasc/supplement/utils.py
javierggt: Fixed

javierggt:
I just added changes according to all these comments. I hope this is it, because I rebase the branch in #86 in my local copy, fixed conflicts, and started doing some refactoring.

jeanconn:
Sure.  I had been thinking one table would be fine, but two is also fine.

agasc/scripts/update_obs_status.py
```diff
@@ -18,26 +18,32 @@
...
     """
+    suppl_file = Path(suppl_file)
+
     if not bad_star_ids:
         logger.info('Nothing to update')
         return
 
-    if not Path(suppl_file).exists():
+    if not suppl_file.exists():
+        if not create:
```
jeanconn, line 44: I'm if we have a create flag, the logger message might not be "warning".  But that's a nit.

jeanconn (approved):

agasc/scripts/update_obs_status.py
javierggt: that's right. I didn't think of that. I just left the message that was there (which I had made inconsistent because it said "creating new file" and then raised an exception)

agasc/supplement/utils.py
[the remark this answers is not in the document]
taldcroft: Ah right I see. Never mind.

taldcroft (approved):
Looks good, thanks!!

taldcroft:
@javierggt - I'm leaving to you to merge just to be sure.

javierggt:
Hey @taldcroft,

I had changed the call to Table.write as you suggested, but now I get this error which results from passing a pathlib.Path instead of a string:
```
In [1]: from pathlib import Path 
   ...: from astropy.table import Table 
   ...: t = Table([{'a':1, 'b':2}]) 
   ...: filename = Path('whatever.h5') 
   ...: t.write(filename, format='hdf5', path='versions', append=True, overwrite=True) 
   ...:                                                                                                                                                      
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-1045fcc97cd7> in <module>
      3 t = Table([{'a':1, 'b':2}])
      4 filename = Path('whatever.h5')
----> 5 t.write(filename, format='hdf5', path='versions', append=True, overwrite=True)

~/SAO/miniconda3-shiny/envs/ska3-flight-2021.2rc4/lib/python3.8/site-packages/astropy/table/connect.py in __call__(self, serialize_method, *args, **kwargs)
    125         instance = self._instance
    126         with serialize_method_as(instance, serialize_method):
--> 127             registry.write(instance, *args, **kwargs)

~/SAO/miniconda3-shiny/envs/ska3-flight-2021.2rc4/lib/python3.8/site-packages/astropy/io/registry.py in write(data, format, *args, **kwargs)
    561 
    562     writer = get_writer(format, data.__class__)
--> 563     writer(data, *args, **kwargs)
    564 
    565 

~/SAO/miniconda3-shiny/envs/ska3-flight-2021.2rc4/lib/python3.8/site-packages/astropy/io/misc/hdf5.py in write_table_hdf5(table, output, path, compression, append, overwrite, serialize_meta, **create_dataset_kwargs)
    314     else:
    315 
--> 316         raise TypeError('output should be a string or an h5py File or '
    317                         'Group object')
    318 

TypeError: output should be a string or an h5py File or Group object
```

When you told me, I tried, it worked and I though "ah, well, maybe I got it wrong"... but no.

taldcroft:
It looks like a bug in astropy, but if you remove the `format='hdf5'` then it works. Go figure. Or you can put the `str()` back in, sorry for the noise.

taldcroft:
I checked and this is still a problem in astropy master. @javierggt - can you file a bug report?

PR #86, opened 2021-02-12
Added a mechanism to set mag_aca and mag_aca_err manually in the supplement, and a unique star-observation ID

### LINKED ISSUE #69: Need mechanism to set mag and mag_err manually
This example shows a case of a star that is clearly much fainter than the catalog mag, at least 12th mag or fainter. If it was brighter than that it would have gotten found at least a few times in the repeated guide searches. We have talked about just marking this star `bad`, but this is not really the right thing because proseco and starcheck will still consider it to be 9.91 mag for spoiling and plotting.

Since none of our tools has the concept of a lower limit for mag, we need to do *something*. For instance just assign it to be 12.0 +/- 0.1 mag and mark those two star-obs as `bad`. For a case like this the `bad` star list seems inappropriate - the only issue with the star is that the catalog mag was incorrect, but now we have a robust way of encapsulating that information. 

Likewise a star that is too bright should be treated accordingly, by putting our best estimate of the mag into the supplement. Whether we remove the historical cases of this from the `bad` table makes no practical difference.

image

### LINKED ISSUE #80: Obs status update does not check for existing entry
```
(ska3-agasc) ➜  agasc git:(logging) python -m agasc.scripts.update_obs_status --obsid 46686 --agasc-id=175379112 --status=bad --comments="ACA high background event"
2021-02-01 06:15:21,080 updating "obs" table in agasc_supplement.h5
2021-02-01 06:15:21,080 Appending agasc_id='175379112', obsid='46686', status={'ok': 0, 'comments': 'ACA high background event'}
(ska3-agasc) ➜  agasc git:(logging) 
(ska3-agasc) ➜  agasc git:(logging) ipythonPython 3.8.3 (default, Jul  2 2020, 11:26:31) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.16.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import tables                                                                                                                           

In [2]: h5 = tables.open_file('agasc_supplement.h5')                                                                                            

In [3]: h5.root.obs[:]                                                                                                                          
Out[3]: 
array([(46686, 175379112, 0, b'ACA high background event'),
       (46686, 175379112, 0, b'ACA high background event')],
      dtype=[('obsid', '<i4'), ('agasc_id', '<i4'), ('ok', '<i4'), ('comments', 'S80')])

In [4]:                                                                                                                                         
Do you really want to exit ([y]/n)? 
Closing remaining open files:agasc_supplement.h5...done
(ska3-agasc) ➜  agasc git:(logging) python -m agasc.scripts.update_obs_status --obsid 46686 --agasc-id=175379112 --status=ok --comments="differene" 
2021-02-01 06:25:10,750 updating "obs" table in agasc_supplement.h5
2021-02-01 06:25:10,750 Appending agasc_id='175379112', obsid='46686', status={'ok': 1, 'comments': 'differene'}
(ska3-agasc) ➜  agasc git:(logging) ipythonPython 3.8.3 (default, Jul  2 2020, 11:26:31) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.16.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import tables                                                                                                                           

In [2]: tables.open_file('agasc_supplement.h5').root.obs[:]                                                                                     
Out[2]: 
array([(46686, 175379112, 0, b'ACA high background event'),
       (46686, 175379112, 1, b'differene')],
      dtype=[('obsid', '<i4'), ('agasc_id', '<i4'), ('ok', '<i4'), ('comments', 'S80')])
```

### LINKED ISSUE #81: Add obs/dwell date to 'obs' table in supplement
Add obs/dwell date to 'obs' table in supplement

This should help us define dwells uniquely.

javierggt:
## Description

This PR includes the following changes:
- Issue #81: Changes the obs table to include an observation ID, which is the mp_starcat_time. Observations are then identified by (observation_id, agasc_id) instead of (obsid, agasc_id). `mp_starcat_time` is still used within `star_obs_catalogs.py`.
- introduces options to the `agasc-obs-status` script to change `mag_aca` and `mag_aca_err`, as requested in issue #69.
- refactors the functions to update tables in the supplement:
  - relocating functions to the `agasc.supplement.utils` module
  - renamed some scripts:
    - `update_obs_status` -> `update_supplement`
    - `agasc-mag-estimate-report` -> `agasc-magnitudes-report`
    - `agasc-supplement-update` -> `agasc-update-magnitudes`
    - `agasc-supplement-obs-status` -> `agasc-update-supplement`
    - `agasc-supplement-bad-star` was removed, functionality is in `agasc-update-supplement`
  - the refactoring causes a few changes in behavior:
    - when an observation appears more than once in the `obs` section, the last appearance is kept (the behavior depended on whether it was repeated in the yaml file or in the command line). This is consistent with yaml.
    - when one tries to update an observation that is not in the catalog, an exception is raised (it used to be silently ignored).
- fixes a bug in the test, where some checks were not run (and they should have failed). A fixed that and introduced changes so it does not happen again.

Comments:
- The current implementation still requires last_obs_date to be given
- some testing is still missing

## Testing

- [x] Functional testing is like in #83:
   ```
   export AGASC_DIR=`pwd`
   cp $SKA/data/agasc/* .;
   #cp ~/SAO/ska/data/agasc/miniagasc* .;

   # create a list of interesting stars to test
   echo 331751984 > agasc_id # an OK new star
   echo 991559968 >> agasc_id # star fails (anyway there should be a single-star page for this star)
   echo 505545160 >> agasc_id # only one suspect obs (anyway there should be a singla-star page for this star)
   echo 42864776 >> agasc_id # an OK older star that gets updated
   echo 1200760024 >> agasc_id  # one suspect OBS (the suspect OBS should show up in the single-star page)
   echo 314189096 >> agasc_id
   echo 553396016 >> agasc_id

   # start with a set of OK stars
   agasc-update-magnitudes --log-level debug --start 2011:151 --stop 2011:152 --report
   agasc-update-magnitudes --log-level debug --start 2019:280 --stop 2019:281 --report

   # add the interesting stars, check the report and check the resulting obs_status.yml
   agasc-update-magnitudes --agasc-id-file agasc_id --log-level debug --report

   # check obs_status.yml, which should include observations, including the observation ID (the date)

   # One could update the supplement 'obs' table with this...
   # agasc-update-supplement --obs-status-file obs_status.yml # and check supplement

   # but that would not update the existing report. We do the following instead
   # the important thing is that the report directory must be the same, otherwise it does not work
   # the report directory is given by the stop time (this still needs a better approach)
   agasc-update-magnitudes --obs-status-file obs_status.yml --log-level debug --report
   # and check supplement
   ```

Fixes #69 
Fixes #80
Fixes #81


taldcroft:
One slight corner case is if we want to override the mag for a star that has never been purposely observed but instead showed up as an acquisition spoiler star that is brighter than a companion bona fide ACQ star.

javierggt:
> One slight corner case is if we want to override the mag for a star that has never been purposely observed but instead showed up as an acquisition spoiler star that is brighter than a companion bona fide ACQ star.

ah! That is one case I did not know about. So then should I allow stars that have never been observed?

taldcroft:
> ah! That is one case I did not know about. So then should I allow stars that have never been observed?

These stars will have been "observed", but not on purpose. So yes, we need a way to capture the observed mag.

taldcroft:
@javierggt - what's the status on this? It is still shown as a WIP and doesn't have functional testing shown.

javierggt:
> what's the status on this?

This currently includes the changes this PR was intended for. I just pushed after rebasing on master. However... I do not want to have two highly-correlated PRs again, and issue #81 would touch on the same places, which means functional testing has to be repeated and tests need to be updated. I would rather do it once (i.e.: closing #81 in this PR).

I expect to have this PR ready today.

javierggt:
One option would be to merge this right now, with functional tests in the next PR.

agasc/supplement/utils.py
```diff
@@ -134,3 +144,191 @@ def save_version(filename, table_name):
...
+    table = unique(table, keys=keys, keep='first')
+
+    # for now I require no masked elements.
+    if table.mask is not None and np.any(table.mask):
+        masked = [name for name in table.colnames if np.any(table[name].mask)]
+        raise Exception('"{path}" table should have no masked elements, '
+                        'but element in these columns are masked: ',
+                        ', '.join(masked))
+
+    # entries already in supplement
```
taldcroft, line 188: I guess this works OK here because of the sizes, but in general I'm worried about this approach as being `O(N*M)` in operations and memory. I think an equivalent approach that is `O(N + M)` would be using a dict to make an index e.g (but this is untested of course).
```
index_suppl = {key: i for i, key in enumerate(suppl_table.iterrows(*keys))}

i = []
j = []
k = []
for i_table, key in enumerate(table.iterrows(*keys)):
    try:
        i.append(index_suppl[key])
        j.append(i_table)
    except KeyError:
        k.append(i_table)
```

agasc/supplement/utils.py
```diff
@@ -134,3 +144,191 @@ def save_version(filename, table_name):
...
+        logger.info('dry run, not saving anything')
+
+
+def get_table(filename, path, dtype=None, create=False):
+    """
+    Gets a table from an HDF5 file. Optionally creates the file if it does not exist.
+    Creates the table if it does not exist.
+
+
+    :param filename:
```
taldcroft, line 223: Missing the rest of the docstring.

agasc/supplement/utils.py
```diff
@@ -134,3 +144,191 @@ def save_version(filename, table_name):
...
+    :param path: str
+        the path of the table in the supplement ('obs', 'mags')
```
taldcroft, line 225: Capitalize and in general do not start with `the`.

agasc/supplement/utils.py
```diff
@@ -134,3 +144,191 @@ def save_version(filename, table_name):
...
+
+    # for now I require no masked elements.
+    if table.mask is not None and np.any(table.mask):
+        masked = [name for name in table.colnames if np.any(table[name].mask)]
+        raise Exception('"{path}" table should have no masked elements, '
+                        'but element in these columns are masked: ',
+                        ', '.join(masked))
+
+    # entries already in supplement
+    intersect = (table[keys[0]][None] == suppl_table[keys[0]][:, None])
```
taldcroft, line 189: To make this approach more obvious/readable  please use `table[keys[0]][None, :]`.

agasc/supplement/utils.py
```diff
@@ -134,3 +144,191 @@ def save_version(filename, table_name):
...
+    if not dry_run:
+        suppl_table.write(str(filename), format='hdf5', path=path, append=True, overwrite=True)
+        save_version(filename, path)
+    else:
+        logger.info('dry run, not saving anything')
+
+
+def get_table(filename, path, dtype=None, create=False):
+    """
+    Gets a table from an HDF5 file. Optionally creates the file if it does not exist.
```
taldcroft, line 219: This comment is confusing because this function does not actually create a new file, instead it only issues the warning that the calling code will later create that file.

agasc/supplement/utils.py
```diff
@@ -134,3 +144,191 @@ def save_version(filename, table_name):
...
+    :param create: bool
+        Create a supplement file if it does not exist
+    """
+    if not Path(filename).exists():
+        if not create:
+            raise FileExistsError(filename)
+        logger.warning(f'Creating a new AGASC supplement: {filename}')
+
+    try:
+        result = Table.read(filename, format='hdf5', path=path)
```
taldcroft, line 239: Probably needs `str(filename)` if `filename` is a `Path`.

agasc/supplement/utils.py
```diff
@@ -134,3 +144,191 @@ def save_version(filename, table_name):
...
+            raise
+
+    return result
+
+
+def add_bad_star(filename, bad, dry_run=False, create=False):
+    """
+    Update the 'bad' table of the AGASC supplement.
+
+    :param filename: pathlib.Path
```
taldcroft, line 253: Presumably this can also be `str`.  This and the rest of these args need descriptions. (This comment applies to several other functions as well).

taldcroft (changes requested):

agasc/supplement/utils.py
```diff
@@ -20,6 +22,26 @@
 AGASC_SUPPLEMENT_TABLES = ('mags', 'bad', 'obs', 'last_updated', 'agasc_versions')
```
jeanconn, line 22: I think 'obs' is now 'obs_status' .

jeanconn: No wait, did it go the other way?  I'll review.

agasc/supplement/magnitudes/update_mag_supplement.py
```diff
@@ -277,18 +269,7 @@ def update_supplement(agasc_stats, filename, include_all=True):
...
-        new_stars = outliers_new['agasc_id']
-        updated_stars = np.array([], dtype=mags_dtype)
-
-    mode = 'r+' if filename.exists() else 'w'
-    with tables.File(filename, mode) as h5:
-        if 'mags' in h5.root:
-            h5.remove_node('/mags')
-        h5.create_table('/', 'mags', outliers)
-    save_version(filename, 'mags')
+    update_mags_table(outliers)
```
jeanconn, line 272: This looks not to match the current function definition for update_mags_table?

agasc/supplement/utils.py
javierggt: This should not be O(N*M) in terms of memory, because numpy does not actually create the broadcasted arrays. It only reshapes by adding an extra dimension of length 1.

It will be O(N*M) in terms of computation, because it compares all pairs.

In our case, I don't think it matters at all.

agasc/supplement/magnitudes/update_mag_supplement.py
javierggt: That is correct. I am actually undoing this change, because one does not gain much by calling the update function. The update itself is different.

agasc/supplement/utils.py
```diff
@@ -134,3 +144,191 @@ def save_version(filename, table_name):
                    append=True, overwrite=True)
     last_updated.write(str(filename), format='hdf5', path='last_updated',
                        append=True, overwrite=True)
+
+
+def update_table(filename, table, path, dtype, keys=['agasc_id'], dry_run=False, create=False):
```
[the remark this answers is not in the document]
javierggt: I just removed the default.

agasc/supplement/utils.py
javierggt: Let me know if you want me to change this, because I would leave it the way it is.

agasc/supplement/utils.py
javierggt: I don't know if you just mean there should be no extra space. I edited the docstring anyway.

agasc/supplement/utils.py
javierggt: Done.

agasc/supplement/utils.py
javierggt: Done.

agasc/supplement/utils.py
javierggt: Right. I edited the docstring, and also made this function private. I changed logging message, but still, this is a specific function, not general.

agasc/supplement/utils.py
javierggt: Just fixed at the top of the function.

agasc/supplement/utils.py
javierggt: I improved the docstrings a bit.

jeanconn:
I wanted to "work along from home" with these examples, but I'm having trouble making a supplement that I can then update manually.  I tried from the "flight" supplement (I think) and just get:

---
  968 OBSIDs,  34 stars,  0 failed stars,  0 failed observations,  0 global errors
2021-02-19 17:09:11,502 Updating test_agasc/mag_stats_agasc.fits
2021-02-19 17:09:12,007 Updating test_agasc/mag_stats_obsid.fits
Traceback (most recent call last):
  File "/fido.real/conda/envs/agasc/bin/agasc-update-magnitudes", line 33, in <module>
    sys.exit(load_entry_point('agasc', 'console_scripts', 'agasc-update-magnitudes')())
  File "/proj/sot/ska/jeanproj/git/agasc/agasc/scripts/update_mag_supplement.py", line 113, in main
    update_mag_supplement.do(
  File "/proj/sot/ska/jeanproj/git/agasc/agasc/supplement/magnitudes/update_mag_supplement.py", line 476, in do
    update_mag_stats(obs_stats, agasc_stats, fails, output_dir)
  File "/proj/sot/ska/jeanproj/git/agasc/agasc/supplement/magnitudes/update_mag_supplement.py", line 196, in update_mag_stats
    obs_stats = _update_table(table.Table.read(filename), obs_stats,
  File "/proj/sot/ska/jeanproj/git/agasc/agasc/supplement/magnitudes/update_mag_supplement.py", line 157, in _update_table
    assert table_old.as_array().dtype.names == table_new.as_array().dtype.names, \
AssertionError: Tables have different dtype
---

The rc supplement had (last time I looked) the old table names.  Not sure which process is validated for starting out here.

javierggt:
this process validates starting from scratch, with a supplement with no tables. It did not start with the existing supplement.

jeanconn:
So, will we need an updated process to avoid throwing away our current bad list?

javierggt:
> So, will we need an updated process to avoid throwing away our current bad list?

sorry, when I said "from scratch" I should have said "from current flight", with no "new" tables (mags/obs/version...)

jeanconn:
OK.  I think my bug report text earlier in the thread was "from current flight" so I'll retry if that should be working from master or this branch.

agasc/supplement/utils.py
taldcroft: The `intersect` array is N*M, no? 

You can leave this here, but in general for this kind of algorithm which is roughly set intersections you should look for an O(N+M) approach. The first time you put in a couple of 1e6 arrays things will go badly wrong.

agasc/supplement/utils.py
taldcroft: I'm still not seeing the types for `filename` nor a description.

taldcroft (approved):
Just one minor nit on a docstring that you can fix or ignore. I think this is ready to merge so we can continue forward progress.

agasc/supplement/utils.py
javierggt: > The intersect array is N*M, no?

I don't think so. if you do `a[:, None]`, the size of the array in memory does not change, only the shape changes. Then when you call `np.argwhere` it does not return an `N*M` array, so numpy should be able to deal with this without allocating more memory.

agasc/supplement/utils.py
javierggt: Committing them in a minute.

agasc/supplement/utils.py
javierggt: actually you are right, the intersect array is `N*M`.

PR #92, opened 2021-02-22
Added a mechanism to log command-line arguments to be able to re-run

### LINKED ISSUE #90: Specifying report location when re-running agasc-update-magnitudes
Currently, `agasc-update-magnitudes` takes a `--reports-dir` argument. Initially I intended this to be the top-level directory for reports, and the actual subdirectory is set automatically using the date. This is ok for weekly runs.

When I re-run the script using the `--obs-status-file obs_status.yml` argument, the date is not specified anymore (currently --obs-status-file and `--start/stop` are not used together), but the script needs to find the right report directory in order to update it.

Two ways to fix it:
- add a `--reports-subdir` option
- make `--obs-status-file` and `--start/stop` work together and stipulate that when re-running one has to pass the same arguments. That might actually be needed anyway if the `star_obs_catalogs.STARS_OBS` changed since the time when the script was run (i.e.: some stars in the run were observed after the script was run)

javierggt:
## Description

This PR adds the possibility to save a YAML file (we could change it to json, if people prefer...) to log command-line arguments.

When `--start/stop` are not given, we want to first set the arguments and then save, so the next time one runs the script these are explicit, and not determined according to the new date. Some arguments used to be modified inside `agasc.supplement.magnitudes.update_mag_supplement.do`, and they were moved so they are modified before saving the arguments to file.

 I also modified the arguments and docstrings of `agasc.supplement.magnitudes.update_mag_supplement.do`

## Testing

- [x] Functional testing.

I am using [AGASC ID 553396016](https://cxc.cfa.harvard.edu/mta/ASPECT/agasc/supplement_reports/stars/055/553396016/index.html) to test. This star has a suspect observation and has been observed multiple times since then.
```
# starting from and empty directory
export AGASC_DIR=`pwd`
cp $SKA/data/agasc/* .;

# this day includes an observation of AGASC ID 553396016
agasc-update-magnitudes --log-level debug --start 2019:362 --stop 2019:363 --report

# the following has a suspect observation of AGASC ID 553396016
agasc-update-magnitudes --log-level debug --start 2020:096 --stop 2020:097 --report

# this should add the suspect observation to the supplement so it is not counted as a failure
agasc-update-magnitudes --obs-status-file obs_status.yml --args-file supplement_reports/weekly/latest/call_args.yml

# we can check that the observation is now yellow in the dashboard
# AND that there are not observations after 2020:097
```

Fixes #90

### DIFF
--- a/agasc/scripts/update_mag_supplement.py
+++ b/agasc/scripts/update_mag_supplement.py
@@ -9,10 +9,13 @@ import os
 from pathlib import Path
 import argparse
 import logging
+import yaml
 import pyyaks.logger
+
 from agasc.supplement.magnitudes import star_obs_catalogs
 from agasc.supplement.magnitudes import update_mag_supplement
 from agasc.scripts import update_supplement
+from cxotime import CxoTime, units as u
 
 
 def get_parser():
@@ -59,6 +62,9 @@ def get_parser():
     other.add_argument('--no-progress', dest='no_progress',
                        help='Do not show a progress bar',
                        action='store_true')  # this has no default, it will be None.
+    other.add_argument('--args-file',
+                       help='YAML file with arguments to '
+                            'agasc.supplement.magnitudes.update_mag_supplement.do')
     other.add_argument("--dry-run",
                        action="store_true",
                        help="Dry run (no actual file or database updates)")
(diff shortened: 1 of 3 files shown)


taldcroft:
You should be able to merge #86 now and rebase this.

taldcroft (approved):

PR #97, opened 2021-02-28
Obs status initial dispositions (DO NOT MERGE)

taldcroft:
## Description

This is a temporary PR for reviewing and merging obs-status dispositions over the mission, as submitted by each of the three ACA reviewers.

TA has done most of the AGASC IDs ending in `0`.  See 2319043 for the dispositions. I have not done it yet, but I think each of the stars that just has a lower limit for the magnitude (set to MAXMAG) should be included in the `bad` stars table.

### DIFF
--- a/obs_status.yml
+++ b/obs_status.yml
@@ -1,0 +1,870 @@
+obs:
+  - observation_id: 2007:332:13:03:23.621
+    status: 1
+    agasc_id: [5115672]
+    comments: Suspect observation
+  - observation_id: 2013:323:16:25:05.901
+    status: 1
+    agasc_id: [5115672]
+    comments: Suspect observation
+  - observation_id: 2005:056:20:11:01.041
+    status: 1
+    agasc_id: [35263288]
+    comments: Suspect observation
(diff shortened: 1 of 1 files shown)


jeanconn:
It is doable to go through this list and finish up, but I think that we should code in some rules of thumb regarding the "if nothing really tracked use something like a faint limit" for the yaml suggestions.  Otherwise are we supposed to look at the image data for each "Suspect observation" and decide if the "sometimes tracking thing" was a star (so use 'obs' and use the magnitude estimate from processing) or warm/hot CCD pixels (don't use 'obs' and manually set a magnitude something like MAXMAG or half MAXMAG)?  

I'd be inclined to generalize more in the yaml suggestion code at this point and go with "if *something* was tracked poorly (I assume there's a threshold in there already) and there's a magnitude estimate: use the estimate and the obs, add star to bad list.
If nothing tracked for the obsid/agasc_id: set a mag estimate based on MAXMAG, add star to bad list, (not sure what to do in 'obs' table).  What are thoughts?

javierggt:
Closing this as it is superseded by #111

PR #100, opened 2021-03-01
Fix "telemetry mismatch" error

### LINKED ISSUE #70: Resolve the "Mismatch in telemetry between aca_l0 and cheta" failures
As @jeanconn nicely documented, most (or all?) of those cases are related to a safing action that resulted in a Kalman dwell being interrupted.

- The star-obs where these occur is not included in the report itself, I guess because of the failure.

Would it be possible to just use the available intersection of L0 images and cheta telemetry instead of raising an exception?
```
    if len(telem['AOACASEQ']) != len(telem['IMGSIZE']):
        raise MagStatsException(
            "Mismatch in telemetry between aca_l0 and cheta",
            agasc_id=obs['agasc_id'], obsid=obs['obsid'], timeline_id=obs['timeline_id']
        )
```
Basically just carry on with whatever is there, and apply the existing criteria for whether a star-obs is OK. In most of the cases of a safing action, there is actually useful ACA image data available.

javierggt:
## Description

This PR fixes an error that was being raised unnecessarily (See issue #70).

Two main changes:
- use `fetch.Msidset` instead of `fetch.MSIDset`
- remove `events.normal_suns` and `events.safe_suns` intervals from slot data coming from `aca_l0`

I did not use the suggestion in issue #70, but I am using kadi events instead. Which one would be better?

## Testing

- [x] Functional testing
   ```
   In [1]: from agasc.supplement.magnitudes import mag_estimate as me, star_obs_catalogs                                                                         

   In [2]: star_obs_catalogs.load()                                                                                                                              

   In [3]: obs = star_obs_catalogs.STARS_OBS[star_obs_catalogs.STARS_OBS['obsid'] == 5806][0]                                                                    

   In [4]: telem = me.get_telemetry(obs)                                                                                                                         

   In [5]: len(telem['times'])                                                                                                                                   
   Out[5]: 9835
   ```

Fixes #70

### DIFF
--- a/agasc/supplement/magnitudes/mag_estimate.py
+++ b/agasc/supplement/magnitudes/mag_estimate.py
@@ -294,7 +294,7 @@ def get_telemetry(obs):
     slot_data = aca_l0.get_slot_data(start, stop, slot=obs['slot'],
                                      centered_8x8=True, columns=slot_data_cols)
 
-    msid = fetch.MSID(f'AOACMAG{slot}', start, stop)
+    msid = fetch.Msid(f'AOACMAG{slot}', start, stop)
     if len(slot_data) == 0:
         raise MagStatsException('No level 0 data',
                                 agasc_id=obs["agasc_id"],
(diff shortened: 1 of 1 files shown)


taldcroft (approved):
Looks good!

PR #103, opened 2021-03-01
Minor fixes

### LINKED ISSUE #79: Command line arg types should be specified or coerced during parsing
Some of the command line arguments are left as strings, leading to unexpected behavior like below where `agasc_id` and `obsid` are shown in logging as string values. Arguments should be either `None` or the correct type at the point of arg parsing.
```
(ska3-agasc) ➜  agasc git:(logging) python -m agasc.scripts.update_obs_status --obsid 46686 --agasc-id=175379112 --status=bad --comments="ACA high background event"
2021-02-01 06:15:21,080 updating "obs" table in agasc_supplement.h5
2021-02-01 06:15:21,080 Appending agasc_id='175379112', obsid='46686', status={'ok': 0, 'comments': 'ACA high background event'}
```

### LINKED ISSUE #87: Follow-on issue for #83: last obs date 1997:365
After following the functional test steps, these last obs dates look like they represent a default instead of the correct value. Even with no data the star was still "observed".
image

### LINKED ISSUE #88: Follow-on issue for #83: nan catalog mag
The catalog mag is listed as nan, which is not right.
image

### LINKED ISSUE #93: Ignore saturated pixel in supplement processing
The ACA saturated pixel flag is (in practice) an indicator of a too-bright star, not a physical defect in the CCD.  A specific example is 693243160 in obsid 58164, which shows no data in the star obs plot but is actually a too-bright star that is being re-acquired every 8 seconds and then rejected.

For the supplement it is important to use the available star image data to estimate the star mag. For stars with saturated pixels this will under-estimate the true brightness, but that is OK.

### LINKED ISSUE #98: Put the image slot somewhere in the detail report page
I lost some time this morning because I accidentally got the wrong slot from mica for a star.

### LINKED ISSUE #99: Fix bug in masking
Now I understand this. There is a bug in the code:
```
    img_sub.mask *= MASK['mouse_bit']
```
The masked elements are `True`, so this should be (with a reminder to never use arithmetic operations on a boolean value):
```
    img_sub.mask |= MASK['mouse_bit']
```
So all of the mags are being computed with the full 8x8 image, so the bias offset effect is much larger due to the 28 edge pixels in the 8x8 readout. After fixing this bug the computed mag is 12.7, which is close enough to 12.0 that we may not need to do the bias offset computation.

### LINKED ISSUE #106: Need to include `use_supplement=False` in supplement processing code
Parts of the supplement processing query the AGASC using `agasc.get_star` in order to get native catalog information (e.g. ACA_MAG). These calls needs to include `use_supplement=False` to avoid getting the estimated mag from the supplement.

### LINKED ISSUE #108: Followup to the "mismatch in telemetry between aca_l0 and cheta" issue
It turns out that issue #70 was not the whole story...

In the current implementation, I first query for a single MSID ('AOACMAG{slot}') and use that to set the times byt finding the intersection of `msid.times` with the times in `slot_data` from `mica.archive.l0`. That does not always work... e.g. OBSID 50255, agasc_id 991559968.

In the changes for issue #70, I started using `fetch.Msid` and `fetch.Msidset` instead of `fetch.MSID` and `fetch.MSIDset`. It took me a while to realize that the `bads` attribute is different depending on the **other** MSIDs in the set. For example, the following, the times for AOACMAG3, which I am using to define the times, are different when fetched together with AOACASEQ:
```
In [2]: msid = fetch.Msid('AOACMAG3', 605480583, 605480588)                                                                                                                               

In [3]: msid.times - 605480583.76560163                                                                                                                                                   
Out[3]: array([0.        , 1.0250001 , 2.05000007, 3.07500017, 4.10000026])

In [4]: msids = fetch.Msidset(['AOACASEQ', 'AOACMAG3'], 605480583, 605480588)                                                                                                             

In [5]: msids['AOACMAG3'].times - 605480583.76560163                                                                                                                                      
Out[5]: array([0.        , 1.0250001 , 3.07500017, 4.10000026])
```

as a result of this, `t=605480585.8156017` is missing from cheta telemetry but present in l0 slot data.

----->  "mismatch in telemetry between aca_l0 and cheta"

javierggt:
## Description

This PR introduces some small changes:
- specify "use_supplement=False" when calling agasc.get_star in mag_estimate (#106)
- small changes in error loggging
- specifying argument types (#79)
- setting default values in agasc_stats (#87, #88 )
- stop using the saturated pixel flag to select times (#93)
- Display the slot number somewhere in the report (#98)
- fixed an error combining masks (#99)
- removed the "mismatch in telemetry between aca_l0 and cheta" exception, which happened when the number of samples from `cheta.fetch` differed from the number of samples from `mica.archive.l0.get_slot_data`. Now the times are the intersection of these times by construction. (#108)

## Testing

- [ ] Functional testing:
   - For #93 and #99, I only checked it runs and the magnitude in the end seems reasonable.
   - To check for #79, run this and verify the resulting supplement
      ```
      python -m agasc.scripts.update_supplement --obsid 46686 --agasc-id=175379112 --status=bad --comments="ACA high background event" --log-level=debug
      ```
   - To check for #87 #88, and #98
      ```
      export AGASC_DIR=`pwd`
      cp ~/SAO/ska/data/agasc/* .
      echo 505545160 > agasc_id
      echo 991559968 >> agasc_id
      agasc-update-magnitudes --agasc-id-file agasc_id --log-level debug --report
      # check that mag_aca and last_obs are set, check that the slot is shown in star report
      ```
   - To check for #108
      ```
      export AGASC_DIR=`pwd`
      cp ~/SAO/ska/data/agasc/* .
      agasc-update-magnitudes --agasc-id-file suspect_agasc_id --log-level debug --report
      # check that a bunch of stars succeded now. The only failures are the ones where all OBS were suspect.
      ```
Fixes #79 
Fixes #87 
Fixes #88
Fixes #93 
Fixes #98
Fixes #99 
Fixes #106
Fixes #108

### DIFF
--- a/agasc/supplement/magnitudes/mag_estimate.py
+++ b/agasc/supplement/magnitudes/mag_estimate.py
@@ -313,7 +313,7 @@ def get_telemetry(obs):
 
     if len(times) == 0:
         # the intersection was null.
-        raise MagStatsException('Time mismatch between cheta and level0',
+        raise MagStatsException('No matching times between cheta and level0',
                                 agasc_id=obs["agasc_id"],
                                 obsid=obs["obsid"],
                                 mp_starcat_time=obs["mp_starcat_time"])
--- a/agasc/supplement/magnitudes/mag_estimate.py
+++ b/agasc/supplement/magnitudes/mag_estimate.py
@@ -363,8 +363,7 @@ def get_telemetry(obs):
     for name in ['AOACIIR', 'AOACISP']:
         telem[name] = np.char.rstrip(telem[name])
     ok = (telem['AOACASEQ'] == 'KALM') & (telem['AOACIIR'] == 'OK') & \
-         (telem['AOACISP'] == 'OK') & (telem['AOPCADMD'] == 'NPNT') & \
-         (telem['AOACFCT'] == 'TRAK')
+         (telem['AOPCADMD'] == 'NPNT') & (telem['AOACFCT'] == 'TRAK')
 
     # etc...
     logger.debug('    Adding magnitude estimates')
@@ -660,8 +659,7 @@ def calc_obs_stats(telem):
     times = telem['times']
 
     kalman = (telem['AOACASEQ'] == 'KALM') & (telem['AOPCADMD'] == 'NPNT')
-    track = (telem['AOACIIR'] == 'OK') & (telem['AOACISP'] == 'OK') & \
-            (telem['AOACFCT'] == 'TRAK')
+    track = (telem['AOACIIR'] == 'OK') & (telem['AOACFCT'] == 'TRAK')
     dr3 = (telem['dr'] < 3)
     dr5 = (telem['dr'] < 5)
 
--- a/agasc/supplement/magnitudes/mag_estimate.py
+++ b/agasc/supplement/magnitudes/mag_estimate.py
@@ -897,9 +897,17 @@ def get_agasc_id_stats(agasc_id, obs_status_override=None, tstop=None):
     stats['mean_corrected'] = np.nan
     stats['weighted_mean'] = np.nan
 
-    result['n_obsids_fail'] = len(failures)
-    result['n_obsids_suspect'] = np.sum(stats['obs_suspect'])
-    result['n_obsids'] = n_obsids
+    star = get_star(agasc_id, date=all_telem['times'][0])
+
+    result.update({
+        'last_obs_time': last_obs_time,
+        'mag_aca': star['MAG_ACA'],
+        'mag_aca_err': star['MAG_ACA_ERR'] / 100,
+        'color': star['COLOR1'],
+        'n_obsids_fail': len(failures),
+        'n_obsids_suspect': np.sum(stats['obs_suspect']),
+        'n_obsids': n_obsids,
+    })
 
     if not np.any(~excluded_obs):
         logger.debug(f'  Skipping star in get_agasc_id_stats({agasc_id=}).'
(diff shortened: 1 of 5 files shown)


agasc/supplement/magnitudes/mag_estimate.py
```diff
@@ -881,7 +881,7 @@ def get_agasc_id_stats(agasc_id, obs_status_override=None, tstop=None):
     stats['mean_corrected'] = np.nan
     stats['weighted_mean'] = np.nan
 
-    star = get_star(agasc_id, date=all_telem['times'][0])
+    star = get_star(agasc_id)
```
taldcroft, line 884: Should this have `use_supplement=False`? Otherwise the `MAG_ACA` below will be from the supplement if available.

jeanconn: I thought we'd captured that as a separate issue but I guess I forgot to do so.

taldcroft:
I looked through each commit and they all look reasonable. Thanks for the atomic commits and helpful commit messages. For 4e98da8, I couldn't easily see what this is doing without opening the code, but if functional testing shows that it closes #87 then 👍 . So this probably just needs basic functional testing.

agasc/supplement/magnitudes/mag_estimate.py
taldcroft: This is now #106

taldcroft (approved):
When you think it's ready.

PR #107, opened 2021-03-04
Default dtype

javierggt:
## Description

This PR makes two changes:
- set default dtypes for supplement tables when they are not found in the file
- properly transform agasc_versions and last_updated tables into dictionaries in get_supplement_table

## Testing

- [x] Functional testing
   - starting from a supplement with no tables:
      ```
      from agasc import get_supplement_table, save_version
      get_supplement_table('mags', as_dict=False), get_supplement_table('agasc_versions', as_dict=True)
      ```
      gives
      ```                                                                                                                                
     /Users/javierg/SAO/git/agasc/agasc/supplement/utils.py:88: UserWarning: No dataset 'mags' in /Users/javierg/SAO/ska/data/agasc/agasc_supplement.h5, returning empty table
       warnings.warn(f"No dataset '{name}' in {supplement_file},"
     /Users/javierg/SAO/git/agasc/agasc/supplement/utils.py:88: UserWarning: No dataset 'agasc_versions' in /Users/javierg/SAO/ska/data/agasc/agasc_supplement.h5, returning empty table
       warnings.warn(f"No dataset '{name}' in {supplement_file},"
     Out[1]: 
     (<Table length=0>
      agasc_id mag_aca mag_aca_err last_obs_time
       int32   float32   float32      float64   
      -------- ------- ----------- -------------,
      {})
     ```
  - starting from a supplement with tables:
     ```
     from agasc import get_supplement_table, save_version
     get_supplement_table('agasc_versions', as_dict=True)
     ```
     gives
     ```
    {'mags': b'4.10.3.dev116+g2319043',
     'supplement': b'4.10.3.dev116+g2319043',
     'obs': b'4.10.3.dev116+g2319043'}

     ```
Fixes #

### DIFF
--- a/agasc/supplement/utils.py
+++ b/agasc/supplement/utils.py
@@ -90,19 +90,22 @@ def get_supplement_table(name, agasc_dir=None, as_dict=False):
             dat = np.array([], dtype=dtypes[name])
 
     if as_dict:
-        out = {}
-        keys_names = {
-            'mags': ['agasc_id'],
-            'bad': ['agasc_id'],
-            'obs': ['agasc_id', 'mp_starcat_time']}
-        key_names = keys_names[name]
-        for row in dat:
-            # Make the key, coercing the values from numpy to native Python
-            key = tuple(row[nm].item() for nm in key_names)
-            if len(key) == 1:
-                key = key[0]
-            # Make the value from the remaining non-key column names
-            out[key] = {nm: row[nm].item() for nm in row.dtype.names if nm not in key_names}
+        if name in ['agasc_versions', 'last_updated']:
+            out = {name: dat[0][name] for name in dat.dtype.names} if len(dat) else {}
+        else:
+            out = {}
+            keys_names = {
+                'mags': ['agasc_id'],
+                'bad': ['agasc_id'],
+                'obs': ['agasc_id', 'mp_starcat_time']}
+            key_names = keys_names[name]
+            for row in dat:
+                # Make the key, coercing the values from numpy to native Python
+                key = tuple(row[nm].item() for nm in key_names)
+                if len(key) == 1:
+                    key = key[0]
+                # Make the value from the remaining non-key column names
+                out[key] = {nm: row[nm].item() for nm in row.dtype.names if nm not in key_names}
     else:
         out = Table(dat)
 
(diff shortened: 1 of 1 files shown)


jeanconn:
I think I'm coming around to the fact that the "empty" supplement with modern tables doesn't have much value for our testing over just the old version of the supplement (with just 'bad' defined).

I note that it looks to me that the newest empty supplement doesn't seem to have an 'obs' table so I'm guessing something wasn't perfect in these changes but maybe we don't care?

jeanconn:
Oh wait, maybe I had a rev-1 version of that file?  I just grabbed the empty supplement file again and see 'obs'.  My comment about the empty supplement file perhaps not helping us remains true.

agasc/supplement/utils.py
```diff
@@ -74,6 +75,8 @@ def get_supplement_table(name, agasc_dir=None, as_dict=False):
     """
     agasc_dir = default_agasc_dir() if agasc_dir is None else Path(agasc_dir)
 
+    dtypes = defaultdict(lambda: [], {'bad': BAD_DTYPE, 'mags': MAGS_DTYPE, 'obs': OBS_DTYPE})
```
taldcroft, line 78: I think this is typically written as `defaultdict(list, ..)`, no?  But more to the point is the even more common thing below to supply the default in a `.get`, as in below where the default is specified in the same line of code.
```
dat = np.array([], dtype=dtypes.get(name, []))
```

agasc/supplement/utils.py
```diff
@@ -84,22 +87,25 @@ def get_supplement_table(name, agasc_dir=None, as_dict=False):
         except tables.NoSuchNodeError:
             warnings.warn(f"No dataset '{name}' in {supplement_file},"
                           " returning empty table")
-            dat = []
+            dat = np.array([], dtype=dtypes[name])
```
taldcroft, line 90: Setting `dtypes=[]` seems weird to me. I would not be 100% sure what that does, except for guessing it is an empty structured array. I think the code is going to do the same thing if you just set it to `None` since it is just testing for a 0-length array.

taldcroft (approved):
There were a couple of things that made it take longer for me to understand the code, doing sort of unusual things that made me have to think. You can change these or not.

PR #111, opened 2021-03-17
Obs status initial dispositions, take 2 (DO NOT MERGE)

javierggt:
## Description

Similar to #97, this is a temporary PR for reviewing and merging obs-status dispositions over the mission, as submitted by each of the three ACA reviewers.

There are about 100 more suspect observations in this version.

TA has done a bunch of the AGASC IDs ending in `0`.  See 2319043 for the original dispositions.

These are some remaining issues we know of:

- [x]  - 4 UNKNOWN star-obs [DISPOSITION - decided to leave unknown]

   - One of these was obsid 9931 which was a "previously-known weird observation.  For the others, I don't know how much more investigation is reasonable (probably none).
 

- [x] - a new source value was added to the bad list for star 458500128. Should this be added? SSAWG? Another value? [DISPOSITION - new source removed due to OBE bad region]

   - This was in obsid 17085 which I is OBE due to discovery of bad region.
  

- [x] - What to do with "Venus" observations that are not explicitly Venus observations?  759958456 and three stars from OBSID 50707 are in this file. Should the others be removed? - [FIXED]

  - We could add take an action to add obsid 53250 to the mica bad/venus list. - DONE
  - I think Obsid 50707 is actually part of the annealing activity. - Also added to mica
  - Overall it wasn't clear if we should remove "otherwise looks OK" star-obs data from the used data for "known bad" observations.  My thought was yes, let's just chuck that data, because I figured the mags might be contaminated by Venus issues or other weirdness even if the star tracking was OK.

- [x] - agasc ID [82972168](https://cxc.cfa.harvard.edu/mta/ASPECT/tmp/review_agasc_reports/stars/008/82972168/index.html) was affected by an imposter. Should it be in the bad list even though we do not know the magnitude? - [DISPOSITION - added to bad list]

  - This is one of the ones that brought up my question, if a magnitude had to be put in manually, should the star be added to the bad list?  Both because the star couldn't be tracked well and because the bad list seems to be where we store the catalog id that labels the kind of mag table update (max or min or manual etc).  Or do we need some labeling in a different product?

- Small open question about when to use MAXMAG or when to just use the mag value from the star-observation data when the observation data is, well, about the same as MAXMAG.

- [x] - Star [500706352](https://cxc.cfa.harvard.edu/mta/ASPECT/tmp/review_agasc_reports/stars/050/500706352/index.html) is in an 18 sample observation with 9 samples that might be the star?  I wasn't sure if we should still have a minimum number of samples.  For this super-short ER it could also be fine to exclude the star-obs data without prejudice. [DISPOSITION - left alone]

- [x] - Star [376724944](https://cxc.cfa.harvard.edu/mta/ASPECT/tmp/review_agasc_reports/stars/037/376724944/index.html) has the comment "Mag 6.5 spoiler 45 arcsec away".  The target is a 6.46 catalog mag star and I think we just have bad position / no-proper-motion for the star.  But this is a judgement call.  I (Jean) made a couple of those judgement calls in my [PR](https://github.com/sot/agasc/pull/117) but not sure if we want a process.  - [DISPOSITION - wiki updated to make proper-motion bad type a position bad type.  Star added to bad list with that bad type]

- [x] - There was a star id in the text file, 405276920, that I couldn't find in the agasc.  I cut it but not sure if I should have looked anyplace else to see where it came in as a typo? [DISPOSITION - no action needed]

- [x] - Star [8395104](https://cxc.cfa.harvard.edu/mta/ASPECT/tmp/review_agasc_reports/stars/000/8395104/index.html) has the comment "Magnitude changes within observation. Eclipsing binary".  So should it be added to the bad list?  We've used it successfully a bunch of times but if we think the large mag variation in the one obsid is inherent to the star it isn't a good star tracking candidate. [DISPOSITION - was a bad region observation, star-obs now excluded]

- [x] - Star [27919752](https://cxc.cfa.harvard.edu/mta/ASPECT/tmp/review_agasc_reports/stars/002/27919752/index.html) has the comment "Mag 9.3 spoiler 26 arcsec away".  The catalog mag is 9.59.  So either the 9.3 mag star is the target or it is a spoiler that is probably always going to spoil this star.  I think that means this gets added to the bad list.  We haven't really generalized handling of spoilers with the manual dispositions. [DISPOSITION - Added to bad list as bad position]

- [x] - Star [723651200](https://cxc.cfa.harvard.edu/mta/ASPECT/tmp/review_agasc_reports/stars/072/723651200/index.html) has an entry in the mag table with no entry in the star-obs dispositions (the star obs data is probably fine?).  So I'm not sure why the entry is there. [DISPOSITION - entry in mags table removed]

- [x] - Star [693243160](https://cxc.cfa.harvard.edu/mta/ASPECT/tmp/review_agasc_reports/stars/069/693243160/index.html) also has an entry in the mags table (4.5) but no entry in star-obs or bads.  It is probably fine (bright star in this case) but the automated process should also have been fine? [DISPOSITION - entry in mags table removed]

- [x] - Star [1130649544](https://cxc.cfa.harvard.edu/mta/ASPECT/tmp/review_agasc_reports/stars/113/1130649544/index.html) is Eta Car.  Right now it has disposition to just set to MAXMAG.  I think it should be on the bad list and might be more appropriate as a too-bright (but I have that guess from the acquisition data). [DISPOSITION - mags table updated with this as a bright star.  Added to bad as bright star]


obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  mp_starcat_time: 2009:023:22:13:56.794
+  status: 1
+- agasc_id:
+  - 639635448
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2009:024:04:12:45.955
+  status: 1
+- agasc_id:
+  - 395055600
+  comments: UNKNOWN: Never acquired. Unknown issue with observation. Presume star is OK.
```
jeanconn, line 775: These comments with ":" colons in the message don't seem to parse for me with the yaml loader.

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  350392600: 11
+  373564272: 9
+  377102496: 9
+  391120072: 9
+  396501520: 9
+  405276920: 9
+  417611952: 11
+  444743456: 11
+  450103168: 9
+  458500128: 13  # new value
```
jeanconn, line 53: Where is this new value defined?

javierggt: It has not been defined. This was not added because of the magnitude being an upper/lower limit, which are the values known up to now. This star would need your review.

jeanconn: OK, so you wanted to add this to the bad star list because it has variability not captured in the agasc and didn't have a mag code for that.

So should we just add a mag entry with mean mag and an error that captures the variability?  I don't see that it would even really need to be on the bad table but right now the bad table seems to be the only place we have to document our decisions about the values in the manual mags table?

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  mp_starcat_time: 2018:155:22:59:34.754
+  status: 1
+- agasc_id:
+  - 738075032
+  comments: Spoiler 30 arcsec away
+  mp_starcat_time: 2018:158:10:17:36.762
+  status: 1
+- agasc_id:
+  - 49818760
+  comments: Faint star. Affected by imposter. Set mag to MAXMAG
```
jeanconn, line 1915: For a couple of these, I'm not sure if we win much setting to MAXMAG over the observed mag?  This one, bad pixels or not, looks like it has 29% tracked and observed mag 11.03.   Maxmag = 10.859.  But fine this way too.

javierggt: I agree in general (I have not gone to see this specific observation now). Observations that are borderline might have gone one way or another depending on the mood :-)

The question is whether we need to make a better guideline to decide.

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+- agasc_id:
+  - 80616784
+  comments: Tracking affected by imposter
+  mp_starcat_time: 2020:280:11:16:08.609
+  status: 1
+- agasc_id:
+  - 82972168
+  comments: Acquired imposter. Out of view
+  mp_starcat_time: 2020:304:03:02:34.117
+  status: 1
```
jeanconn, line 2107: It isn't clear to me if this one should also be in the bad table.

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  mag_aca: 11.2
+  mag_aca_err: 0.1
+- agasc_id: 92812824
+  mag_aca: 11.141
+  mag_aca_err: 0.1
+- agasc_id: 107881800
+  mag_aca: 10.969
+  mag_aca_err: 0.1
+- agasc_id: 109061768
+  mag_aca: 11.621  # need to check, not from starcheck report
```
jeanconn, line 149: I think the starcheck report at

https://icxc.harvard.edu/mp/mplogs/2019/FEB0419/oflsa/starcheck.html#obsid48456

is probably right, which looks like 11.797.

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2020:055:11:57:17.800
+  status: 1
+- agasc_id:
+  - 33690776
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2020:057:21:03:53.096
+  status: 1
+- agasc_id:
+  - 111028544
```
jeanconn, line 2014: This is one of a handful of stars already in the bad table.  Not sure if we care.

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  mag_aca: 11.25
+  mag_aca_err: 0.1
+- agasc_id: 118498608
+  mag_aca: 12.7
+  mag_aca_err: 0.1
+- agasc_id: 120457376
+  mag_aca: 11.547
+  mag_aca_err: 0.1
+- agasc_id: 144714600
+  mag_aca: 10.097  # need to check, not from starcheck report
```
jeanconn, line 170: https://icxc.harvard.edu/mp/mplogs/2015/AUG1015/oflsa/starcheck.html#obsid51679 suggests 10.672 maxmag

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  status: 1
+- agasc_id:
+  - 373564272
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2009:335:11:29:27.547
+  status: 1
+- agasc_id:
+  - 161357600
+  comments: Almost never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2009:345:18:09:20.636
```
jeanconn, line 801: This seems fine, but I also note there are only 133 samples in this observation.  I think that is enough to determine that the "never got the star" is useful information but I'm not sure of the minimum number of samples required.

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  mag_aca: 11.25
+  mag_aca_err: 0.1
+- agasc_id: 327945264
+  mag_aca: 11.25
+  mag_aca_err: 0.1
+- agasc_id: 328077952
+  mag_aca: 11.016
+  mag_aca_err: 0.1
+- agasc_id: 330967424
+  mag_aca: 10.947  # need to check, not from starcheck report
```
jeanconn, line 236: https://icxc.harvard.edu//mp/mplogs/2006/MAR0606/oflsc/starcheck.html#obsid59087 looks like 11.219

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  mag_aca_err: 0.1
+- agasc_id: 350392600
+  mag_aca: 5.0  # need to check
```
jeanconn, line 239: This one is also probably in a the bad star table (it was in the proseco bad star list) and there we said it was like a 3.6 mag star.  I am concerned that the agasc supplement report doesn't really give clues about it being too bright.

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  mag_aca: 11.297
+  mag_aca_err: 0.1
+- agasc_id: 396501520
+  mag_aca: 10.3
+  mag_aca_err: 0.1
+- agasc_id: 405271520
+  mag_aca: 11.2
+  mag_aca_err: 0.1
+- agasc_id: 405276920
+  mag_aca: 11.225  # need to check, not from starcheck report
```
jeanconn, line 257: The starcheck report suggests 11.203.  Either seems fine to me.

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  mag_aca_err: 0.1
+- agasc_id: 417611952
+  mag_aca: 5.0  # need to check
```
jeanconn, line 260: For the bright ones, 5.0 seems fine.

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  mag_aca: 12.2
+  mag_aca_err: 0.1
+- agasc_id: 497824592
+  mag_aca: 13.078
+  mag_aca_err: 0.1
+- agasc_id: 505545160
+  mag_aca: 10.8
+  mag_aca_err: 0.1
+- agasc_id: 612370792
+  mag_aca: 12.177  # need to check, not from starcheck report
```
jeanconn, line 275: Starcheck has 12.797 .

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  status: 1
+- agasc_id:
+  - 758255832
+  comments: Venus observation
+  mp_starcat_time: 2006:086:08:01:53.448
+  status: 1
+- agasc_id:
+  - 759958456
+  comments: Venus observation
+  mp_starcat_time: 2006:086:08:01:53.448
```
jeanconn, line 581: I thought we were going to exclude all of the Venus star observations?  If so, I think this needs entries for at least 2006:086:11:45:33.448 and 2006:086:13:37:23.448 too?

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  mag_aca: 11.4
+  mag_aca_err: 0.1
+- agasc_id: 647762704
+  mag_aca: 11.453
+  mag_aca_err: 0.1
+- agasc_id: 649070072
+  mag_aca: 12.719
+  mag_aca_err: 0.1
+- agasc_id: 652480464
+  mag_aca: 10.075
```
jeanconn, line 305: I think maxmag should be 11.578

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  mp_starcat_time: 2003:221:17:58:51.959
+  status: 1
+- agasc_id:
+  - 897878136
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2003:228:21:05:34.379
+  status: 1
+- agasc_id:
+  - 857375576
+  comments: Expected star position is >1000 arcsec away during the whole obs
```
jeanconn, line 469: I don't immediately see what was going on in obsid 4469.  The post-facto guide props suggests that we were tracking star 857375848 in slot 6 instead of 857375576.  https://icxc.harvard.edu/aspect/mica_reports/04/04469/props.html Seems fine to just exclude the data for this star-obs combination.

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  mag_aca: 11.0
+  mag_aca_err: 0.1
+- agasc_id: 1092619768
+  mag_aca: 11.288  # need to check, not from starcheck report
+  mag_aca_err: 0.1
+- agasc_id: 1105592952
+  mag_aca: 10.734
+  mag_aca_err: 0.1
+- agasc_id: 1108353064
+  mag_aca: 10.91  # need to check, not from starcheck report
```
jeanconn, line 431: Starcheck has maxmag of 11.094

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+- agasc_id:
+  - 962998672
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2010:142:09:25:14.003
+  status: 1
+- agasc_id:
+  - 904010616
+  comments: Faint star. Near MAXMAG. Marginally acceptable
+  mp_starcat_time: 2010:221:16:22:00.639
+  status: 0
```
jeanconn, line 852: I'm a bit confused here.  If this is status=0 why bother with the manual addition in 'mags'?

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  mag_aca: 11.203
+  mag_aca_err: 0.1
+- agasc_id: 899950912
+  mag_aca: 11.688
+  mag_aca_err: 0.1
+- agasc_id: 901664112
+  mag_aca: 10.406
+  mag_aca_err: 0.1
+- agasc_id: 904010616
+  mag_aca: 10.902  # need to check, not from starcheck report
```
jeanconn, line 398: I note starcheck has maxmag 10.876 but not sure if this entry for 904010616 should stay anyway.  Fine with whatever here.

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  mp_starcat_time: 2016:136:08:42:06.470
+  status: 0
+- agasc_id:
+  - 118359864
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2016:141:00:48:34.306
+  status: 1
+- agasc_id:
+  - 397673760
+  comments: UNKNOWN: SCS 107 at start of observation
```
jeanconn, line 1640: Obsid 18625 went to BSH but sure, SCS107 was probably in there.

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  mp_starcat_time: 2013:313:00:39:05.039
+  status: 1
+- agasc_id:
+  - 898135232
+  comments: Venus observation
+  mp_starcat_time: 2013:313:00:39:05.039
+  status: 1
+- agasc_id:
+  - 898125104
+  comments: Never acquired. Set mag to MAXMAG
```
jeanconn, line 1365: Obsid 53250 was really a Venus observation too, so star 898125104 should be "forgiven".

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+- agasc_id: 879768704
+  mag_aca: 10.859
+  mag_aca_err: 0.1
+- agasc_id: 891159184
+  mag_aca: 11.203
+  mag_aca_err: 0.1
+- agasc_id: 897878136
+  mag_aca: 10.75
+  mag_aca_err: 0.1
+- agasc_id: 898125104
```
jeanconn, line 388: This can be cut.

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  740693864: 9
+  771756656: 9
+  805966472: 9
+  821318496: 9
+  836372744: 9
+  836381648: 9
+  879768704: 9
+  891159184: 9
+  897878136: 9
+  898125104: 9
```
jeanconn, line 88: Can be "forgiven" its Venus failures.

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  mp_starcat_time: 2013:313:00:39:05.039
+  status: 1
+- agasc_id:
+  - 898125104
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2013:313:03:14:15.039
+  status: 1
+- agasc_id:
+  - 898135232
+  comments: Never acquired
```
jeanconn, line 1370: Same comment above about 53250.

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  status: 1
+- agasc_id:
+  - 397680904
+  comments: UNKNOWN: SCS 107 at start of observation
+  mp_starcat_time: 2016:234:07:04:56.469
+  status: 1
+- agasc_id:
+  - 236326704
+  comments: ACA annealing
+  mp_starcat_time: 2016:257:09:35:53.151
```
jeanconn, line 1666: I would think that we'd toss all the guide star data from 50707 and just see three stars here...  But maybe the other stars were used in both catalogs?

obs_status.yml
```diff
@@ -0,0 +1,2147 @@
...
+  status: 1
+- agasc_id:
+  - 694561568
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2020:228:13:19:56.547
+  status: 1
+- agasc_id:
+  - 396501520
+  comments: Never acquired; set mag to MAXMAG
+  mp_starcat_time: 2020:240:12:46:53.718
```
jeanconn, line 2061: I don't see this star in the catalog for 46955 or 46950 so this seems odd?

obs_status.yml
javierggt: I was taking the list of Venus observations from [mica.bad_obsids](https://github.com/sot/mica/blob/master/mica/bad_obsids.py), and obsid 53250 is not there.

I just went to the starcheck page, and it is not marked as a Venus observation. There are other obsids marked as Venus observations in the same page. Am I missing something?

None of the stars you say should be "forgiven" are in Venus observations listed in mica.obsids.

I intend to make a commit removing all Venus star-observations and adding the observations themselves (which will remove other stars not included here). Do you want me to wait until you are done reviewing?

obs_status.yml
javierggt: I wrote "UNKNOWN" because I could not figure out what happened, just that there was the SCS107.

javierggt: so the question is what do you suggest should be the comment?

obs_status.yml
javierggt: What do you mean by "not sure if this entry for 904010616 should stay anyway."?

obs_status.yml
jeanconn: That's really a question about how we will use the comments.

obs_status.yml
jeanconn: Right.  I had just looked up 53250, seen that it is a "Special Case ER" right after a Venus observation and at the same attitude.  It would be a different exercise to see if Venus is still on the CCD.  I assume when I made the list of mica Venus observations I did not look for ERs as I figured we would not dwell on Venus during an ER.

Regarding "There are other obsids marked as Venus observations in the same page. Am I missing something?"  I'm not sure what you mean?

jeanconn: "None of the stars you say should be "forgiven" are in Venus observations listed in mica.obsids."  And did I say this about more than one star?  I'll look again.

obs_status.yml
javierggt: Good catch. If the status is 0, it should not be in 'bad' or 'mags'

obs_status.yml
javierggt: 👍🏼 

Do you suggest modifying the comments?

obs_status.yml
javierggt: What happens is that not all Venus star-observations end up being flagged by the process. Instead of adding star-observations separately, I intend to add the observations themselves, which in turn will remove all star-observations for those OBSIDs.

obs_status.yml
javierggt: 👍🏼

obs_status.yml
jeanconn: Understood and that's what I figured.  It just means that there are Venus star-obs combinations now with accompanying mag information in the current released supplement and the update to add all the Venus star-observations themselves will strip those out of the table.  I *think* that is fine; I just hadn't realized they were already in the data.

obs_status.yml
javierggt: Any star that is in the 'bad' table already should not be here.

All the following are in the obs_status file, but they are already in the bad table, so I will remove them from this file:
- 33690776
- 34996472
- 111028544
- 185871616
- 350392600
- 444743456
- 656409216
- 1117787424
- 1130649544
- 1197635184

(there will be a commit for this)

obs_status.yml
jeanconn: It is a good comment now but not sure if we also want shorthand to say that the star catalog data we have for it must be wrong.

obs_status.yml
javierggt: there is still the question whether we would like the magnitudes for these stars in the mags table, but considering that the magnitude estimate is usually some lower/upper limit, and the star is already marked bad, I do not see the value.

obs_status.yml
javierggt: I am not sure exactly what you are proposing here.

obs_status.yml
javierggt: We do. I left a comment below (reading from bottom to top)

obs_status.yml
jeanconn: I think I was proposing no change; just asking if we had a minimum dwell duration to be included in the table(s).

obs_status.yml
javierggt: what can be cut?

obs_status.yml
javierggt: if they are not here it means they were not classified as "suspect", hence my comment earlier about seeing whether we can change the threshold.

obs_status.yml
javierggt: This is OBSID 46955 and I do see it here: https://icxc.harvard.edu/mp/mplogs/2020/AUG2420/oflsa/starcheck.html#obsid46955

obs_status.yml
```diff
@@ -97,21 +89,12 @@ bad:
...
-  1130649544: 9
   1139171624: 9
   1178221128: 9
-  1197635184: 9
 mags:
 - agasc_id: 5115672
   mag_aca: 11.422
   mag_aca_err: 0.1
-- agasc_id: 33690776
-  mag_aca: 9.984
```
jeanconn: So, this one was already in the bad list, but did it have an entry in the mags table already?  And if not, does it deserve one?

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2020:240:12:46:53.718
+  status: 1
+- agasc_id:
+  - 396501520
+  comments: Never acquired; set mag to MAXMAG
+  mp_starcat_time: 2020:243:11:17:18.150
+  status: 1
+- agasc_id:
+  - 42864912
+  comments: Faint star. Set mag to 12.6 +- 0.3
```
jeanconn, line 1791: Looks like entry in mags is missing?

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2020:280:11:16:08.609
+  status: 1
+- agasc_id:
+  - 82972168
+  comments: Acquired imposter. Out of view
+  mp_starcat_time: 2020:304:03:02:34.117
+  status: 1
+- agasc_id:
+  - 171724672
+  comments: Mag 7.6 spoiler 13 arcsec away
```
jeanconn, line 1831: I think this might be a high proper motion star for which we have no proper motion in the agasc:
http://simbad.u-strasbg.fr/simbad/sim-id?Ident=%40795861&Name=HD%20%2037124&submit=submit ?

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  comments: Almost never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2012:174:04:58:32.901
+  status: 1
+- agasc_id:
+  - 653529952
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2012:187:04:30:17.984
+  status: 1
+- agasc_id:
+  - 740693864
```
jeanconn, line 930: Not sure if obsid 54511 should be added to bad star-obs too (it doesn't look much better than 54506 here).  But probably no effective difference if star added to bad list and mags.

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  status: 1
+- agasc_id:
+  - 1105998760
+  comments: Star gets 2 mags dimmer
+  mp_starcat_time: 2019:190:08:02:57.107
+  status: 1
+- agasc_id:
+  - 836381648
+  comments: Faint star. Affected by imposters. Set mag to MAXMAG
+  mp_starcat_time: 2019:245:18:05:01.870
```
jeanconn, line 1692: Similarly it seems like the 2021:144:07:44:05.596 could be also excluded but shouldn't matter if star gets on bad list.

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mag_aca: 9.047
+  mag_aca_err: 0.1
+- agasc_id: 689441328
+  mag_aca: 12.0
+  mag_aca_err: 0.1
+- agasc_id: 690625640
+  mag_aca: 11.2
+  mag_aca_err: 0.1
+- agasc_id: 693243160
+  mag_aca: 4.5
```
jeanconn, line 303: This looks good to me, but if not in the bad table or the star obs table, it seems like there's no note for "why".  Don't think we care.

obs_status.yml
jeanconn: Gotcha.  This is an ER that ran after an observing-only interrupt, so the obsid was re-used and it gets a little confusing (the mica search by obsid shows a different catalog that is also right but for a different time and not relevant).  So this seems fine.

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2004:200:06:56:32.527
+  status: 1
+- agasc_id:
+  - 920783352
+  comments: NSM at start of observation
+  mp_starcat_time: 2004:200:06:56:32.527
+  status: 1
+- agasc_id:
+  - 897607632
+  comments: Large magnitude variation. Known eclipsing binary
```
jeanconn, line 455: Should we add a larger mag err or put this on the bad star list?

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2004:212:17:56:19.266
+  status: 1
+- agasc_id:
+  - 76813072
+  comments: mag 8.1 spoiler 15 arcsec away
```
jeanconn, line 460: I think the "spoiler" is the star and we have bad coordinates.  So maybe add to bad list?

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mag_aca: 10.719
+  mag_aca_err: 0.1
+- agasc_id: 836372744
+  mag_aca: 11.609
+  mag_aca_err: 0.1
+- agasc_id: 836381648
+  mag_aca: 11.203
+  mag_aca_err: 0.1
+- agasc_id: 879768704
+  mag_aca: 10.859
```
jeanconn, line 348: I see starcheck maxmag of 10.328 (looks like 10.859 is one line up).

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2018:014:04:03:16.301
+  status: 1
+- agasc_id:
+  - 797847184
+  comments: Tracking affected by imposter. Set mag to 12.3 +- 0.2
+  mp_starcat_time: 2018:068:11:54:25.784
+  status: 1
+- agasc_id:
+  - 1105068512
+  comments: spoiler 22 arcsec away
```
jeanconn, line 1611: I suspect this is another high proper motion star http://simbad.u-strasbg.fr/simbad/sim-id?Ident=%403436260&Name=HD%20204961&submit=submit

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2018:069:03:04:54.838
+  status: 1
+- agasc_id:
+  - 872547280
+  comments: Large mag variation. Star is a known eclipsing binary
```
jeanconn, line 1616: Similarly, should we put this on the bad list or add a mag error?

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  status: 1
+- agasc_id:
+  - 615914552
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2005:281:04:44:20.882
+  status: 1
+- agasc_id:
+  - 839385704
+  comments: Almost never acquired. When acquired, it was too bright. Star has been
+    observed other times. Known symbiotic star
```
jeanconn, line 491: Unclear if we should take action to avoid using this as a guide star in the future (bad list or larger mag error).

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2006:220:09:22:11.687
+  status: 1
+- agasc_id:
+  - 879768704
+  comments: Almost never tracked. Set mag to MAXMAG
+  mp_starcat_time: 2007:085:18:25:58.626
+  status: 1
+- agasc_id:
+  - 1020796600
+  comments: Never acquired; set mag 11.5 +/- 0.1
```
jeanconn, line 576: 11.5 is basically MAXMAG

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2017:354:04:35:24.949
+  status: 1
+- agasc_id:
+  - 653529952
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2018:002:23:37:48.551
+  status: 1
+- agasc_id:
+  - 123874048
+  comments: Bright star
```
jeanconn, line 1601: Seems like it should get a mags entry.

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  comments: Star gets 2 mags dimmer
+  mp_starcat_time: 2019:190:08:02:57.107
+  status: 1
+- agasc_id:
+  - 836381648
+  comments: Faint star. Affected by imposters. Set mag to MAXMAG
+  mp_starcat_time: 2019:245:18:05:01.870
+  status: 1
+- agasc_id:
+  - 157951008
```
jeanconn, line 1695: Using this data seems a stretch, but the estimate comes out at about maxmag anyway.

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2019:012:01:07:46.603
+  status: 1
+- agasc_id:
+  - 109061768
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2019:035:04:23:05.219
+  status: 1
+- agasc_id:
+  - 989727248
+  comments: UNKNOWN
```
jeanconn, line 1671: For me, that's a "don't care" because it has a bunch of reasonable observations.

obs_status.yml
javierggt: I think so. This is one of the early ones Tom did before he suggested the MAXMAG solution, and because it does not have the keyword it fell through the cracks.

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2015:234:03:22:43.521
+  status: 1
+- agasc_id:
+  - 376724944
+  comments: Mag 6.5 spoiler 45 arcsec away
+  mp_starcat_time: 2015:331:14:10:26.457
+  status: 1
+- agasc_id:
+  - 195167088
```
jeanconn, line 1325: It looks like some of the other observations are reasonable.  Maybe this doesn't need to be set to maxmag?

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2015:212:20:00:08.125
+  status: 1
+- agasc_id:
+  - 821318496
+  comments: Faint star. Almost never tracked. Set mag to MAXMAG
+  mp_starcat_time: 2015:220:11:07:04.609
+  status: 1
+- agasc_id:
+  - 500706352
+  comments: Failed to track but there are snapshots
```
jeanconn, line 1306: I'm not sure if 9 samples are enough.

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2015:233:22:11:26.142
+  status: 1
+- agasc_id:
+  - 144714600
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2015:234:03:22:43.521
+  status: 1
+- agasc_id:
+  - 376724944
+  comments: Mag 6.5 spoiler 45 arcsec away
```
jeanconn, line 1321: I think this might be the agasc target and the catalog has poor position.

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2009:023:22:13:56.794
+  status: 1
+- agasc_id:
+  - 639635448
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2009:024:04:12:45.955
+  status: 1
+- agasc_id:
+  - 395055600
+  comments: UNKNOWN Never acquired. Unknown issue with observation. Presume star is OK.
```
jeanconn, line 651: There's a web page for obsid 9931 at https://cxc.harvard.edu/mta/ASPECT/obs9931_drops/ so yes, star is probably fine.

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  comments: Never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2009:335:11:29:27.547
+  status: 1
+- agasc_id:
+  - 161357600
+  comments: Almost never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2009:345:18:09:20.636
+  status: 1
+- agasc_id:
+  - 1130649544
```
jeanconn, line 680: This is Eta Car and I think that maxmag reasonable but not sure.

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+- agasc_id:
+  - 612763008
+  comments: Faint star with imposter. Marginally acceptable
+  mp_starcat_time: 2010:285:03:11:03.570
+  status: 0
+- agasc_id:
+  - 612763008
+  comments: Almost never acquired. Faint star
+  mp_starcat_time: 2010:285:03:49:31.385
+  status: 1
```
jeanconn, line 748: I'm not really seeing that the 2 samples of the first observation are more useful than the other observation?  But doesn't seem worth looking at the L0 data either way.

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2011:269:04:58:50.064
+  status: 1
+- agasc_id:
+  - 83631240
+  comments: High background event
+  mp_starcat_time: 2011:275:21:41:33.299
+  status: 1
+- agasc_id:
+  - 764150080
+  comments: Never acquired; set mag 12.0 +/- 0.1
```
jeanconn, line 881: Also add to bad list if needed?

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2013:128:05:54:19.223
+  status: 1
+- agasc_id:
+  - 836372744
+  comments: Almost never acquired. Set mag to MAXMAG
+  mp_starcat_time: 2013:207:12:48:00.907
+  status: 1
+- agasc_id:
+  - 139214712
+  comments: Faint star. Poorly tracked. Set mag to 11.35 +- 0.1
```
jeanconn, line 1051: I don't see this mag actually set.

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+- agasc_id: 693243160
+  mag_aca: 4.5
+  mag_aca_err: 0.1
+- agasc_id: 694561568
+  mag_aca: 9.8
+  mag_aca_err: 0.1
+- agasc_id: 720504096
+  mag_aca: 11.344
+  mag_aca_err: 0.1
+- agasc_id: 723651200
```
jeanconn, line 311: For this one, I suppose I'm not expecting a mags entry on its own?  It looks to me like the star-obs data spits out 12.21 mag.

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+- agasc_id: 377102496
+  mag_aca: 10.797
+  mag_aca_err: 0.1
+- agasc_id: 391120072
+  mag_aca: 11.297
+  mag_aca_err: 0.1
+- agasc_id: 396501520
+  mag_aca: 10.3
+  mag_aca_err: 0.1
+- agasc_id: 405271520
```
jeanconn, line 227: I can't find a record of this star in the agasc.

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2018:081:21:42:04.564
+  status: 0
+- agasc_id:
+  - 899950912
+  comments: Faint star. Affected by hot pixels. Set mag to MAXMAG
+  mp_starcat_time: 2018:155:22:59:34.754
+  status: 1
+- agasc_id:
+  - 738075032
+  comments: Spoiler 30 arcsec away
```
jeanconn, line 1631: I think this is more strictly an imposter.

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  comments: NSM at start of observation
+  mp_starcat_time: 2016:324:12:21:46.079
+  status: 1
+- agasc_id:
+  - 314321208
+  comments: NSM at start of observation
+  mp_starcat_time: 2016:324:12:21:46.079
+  status: 1
+- agasc_id:
+  - 27919752
```
jeanconn, line 1435: Should also be added to bad list if the every SRCH got us back to the spoiler star (or we have bad position for star, either way).

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2019:113:10:12:41.647
+  status: 1
+- agasc_id:
+  - 120457376
+  comments: Faint star affected by imposters. Set magnitude to MAXMAG
+  mp_starcat_time: 2019:121:21:47:55.999
+  status: 1
+- agasc_id:
+  - 1105998760
+  comments: Star gets 2 mags dimmer
```
jeanconn, line 1686: I agree with excluding this observation.  The 2 mags dimmer seems to be in relation to where the star ends up in the dither pattern so not immediately seeing if this a local CCD issue or variant of high background.

obs_status.yml
```diff
@@ -0,0 +1,1995 @@
...
+  mp_starcat_time: 2018:198:01:56:54.706
+  status: 1
+- agasc_id:
+  - 974925328
+  comments: Faint star. Tracking affected by imposters. Marginally acceptable
+  mp_starcat_time: 2018:214:12:04:13.567
+  status: 0
+- agasc_id:
+  - 8395104
+  comments: Magnitude changes within observation. Eclipsing binary
```
jeanconn, line 1651: This one is tricky.  We have a bunch of observations that used this star fine, but do we want to avoid it in the future if it is an eclipsing binary?  Larger mag error or bad star table?

obs_status.yml
jeanconn: I also note that, somewhat similar to the variability seen in star 1105998760 in 48040, the variability here in 458500128 in obsid 17085 seems to be related to the dither pattern.

obs_status.yml
jeanconn: I think maybe this is just a typo of the next one star 405276920 ?

taldcroft:
For the record, here are the dispositions I made for the week of Jun 21. These should be incorporated to this PR or at least not forgotten.
```
obs:
  - mp_starcat_time: 2021:154:16:23:25.445
    obsid: 46314
    status: 1
    agasc_id: [813964024]
    comments: Never acquired. Set mag to MAXMAG
  - mp_starcat_time: 2021:146:20:29:05.330
    obsid: 46330
    status: 1
    agasc_id: [1199706104]
    comments: Never acquired. Set mag to MAXMAG

bad:
  1199706104: 9
  813964024: 9

mags:
  - agasc_id: 1199706104
    mag_aca: 10.547
    mag_aca_err: 0.1
  - agasc_id: 813964024
    mag_aca: 9.812
    mag_aca_err: 0.1
```

obs_status.yml
jeanconn: I've removed this star from the bad list as it looks like the behavior was related to an area on the ccd / camera and not the star.

obs_status.yml
jeanconn: I updated the comments with that note about the catalog.  I think we can just safely exclude these star-obs observations without making code changes to special case them.  It is a separate question on if we should have some fixes in the starcheck database and timelines / kadi to get the catalogs right.

obs_status.yml
jeanconn: There was a star-obs record for this star with status = 0 with a reported mag close to this mag.  So the value of the 'mags' entry was questionable.

obs_status.yml
jeanconn: Was added to bad list.

obs_status.yml
jeanconn: Fixed.

obs_status.yml
jeanconn: I gave it a mags entry.

obs_status.yml
jeanconn: I put this in the bad list as a high proper motion star

obs_status.yml
jeanconn: I added the mag and bad entry.

obs_status.yml
jeanconn: I added this to the bad table as a high proper motion star

obs_status.yml
jeanconn: Doesn't really matter in this case.

PR #115, opened 2021-06-01
Email on supplement promotion

javierggt:
## Description

Added function to send email on supplement promotion, but only if something is actually promoted.

## Testing

- [x] Functional testing on `/proj/sot/ska3/agasc-test`


javierggt:
one question here is what to do with the destination email. Right now it is hardcoded.

PR #120, opened 2021-07-15
More status updates

jeanconn:
## Description

More status updates

## Testing

- [ ] Passes unit tests on MacOS, linux, Windows (at least one required)
- [ ] Functional testing

Fixes #


jeanconn:
@javierggt These seem like all of the remaining actions out of our review.  Any objections to any of these?  I did not re-sort the yaml.

obs_status.yml
```diff
@@ -1445,7 +1446,7 @@ obs:
   status: 1
 - agasc_id:
   - 27919752
-  comments: Mag 9.3 spoiler 26 arcsec away
+  comments: Mag 9.3 spoiler 26 arcsec away.  Set bad with bad position.
```
javierggt, line 1452: strictly speaking, if this is bad position, then it is not a spoiler, so we could remove that, right?

obs_status.yml
```diff
@@ -1331,7 +1332,7 @@ obs:
   status: 1
 - agasc_id:
   - 376724944
-  comments: Mag 6.5 spoiler 45 arcsec away
+  comments: Mag 6.5 spoiler 45 arcsec away. Set bad with bad position.
```
javierggt, line 1337: same comment here: strictly speaking, if this is bad position, then it is not a spoiler, so we could remove that, right?

obs_status.yml
```diff
@@ -1,4 +1,5 @@
 bad:
+  1143744224: 12
```
javierggt, line 7: same comment about the "manual mag 'bad' label" (perhaps it is shown below?)

obs_status.yml
```diff
@@ -1,4 +1,5 @@
 bad:
+  76813072: 10
```
javierggt, line 4: in the other cases of bad position stars, you also changed the comments.

obs_status.yml
```diff
@@ -1,4 +1,5 @@
 bad:
+  1081092600: 12
```
javierggt, line 6: can you remind me why we need a "manual mag 'bad' label"? If the magnitude was set manually it does not mean the star is bad, and instead the star can be used by the selection algorithm based on the magnitude value/error.

javierggt (commented):
I made some comments, and I can live with all these changes.

My only real comment is why to have a 'bad' label for manually-set magnitudes. If the magnitude was set manually, why not let the selection algorithm choose whether to use it or not based on the magnitude? Perhaps we already discussed this. I just don't remember.

obs_status.yml
jeanconn: Right, if the magnitude was set manually it does not **necessarily** mean the star was bad, but in these cases, I thought the magnitude was set manually because the star was poorly tracked and the ACA reviewer needed to make a judgement call.  I think if the star is poorly tracked enough that we need to make that judgement call, it is likely it can also be marked bad with no real downside.

obs_status.yml
jeanconn: Comment updated

jeanconn (commented):
Right, so for the stars that already had "manual" type magnitudes, I did not do much **extra** review to see if they could be converted to "maxmag" estimates or if they really belonged on the bad list.  For the brighter stars that came up (that might be selected as guide stars in the future), that had at least enough of an issue with tracking to get attention and/or get a mag estimate that wasn't awesome, I also added to bad list.  Could go either way, but I think this might be "good enough".

jeanconn:
I also note that I didn't add the manual mag bad label to the wiki list of bad types; it was one of our bad star types before this PR.

javierggt:
> I also note that I didn't add the manual mag bad label to the wiki list of bad types; it was one of our bad star types before this PR.

It might have been me who added it as part of this whole process 😄

javierggt (approved):

PR #124, opened 2021-08-05
remove existing obs_status file if it is not going to be replaced

javierggt:
## Description

Currently, if there are no obs_status changes, nothing is done, and the file from the previous run is left there, which might be confusing. This PR removes and existing obs_status file if it is not going to be replaced.

## Testing

- [ ] Passes unit tests on MacOS, linux, Windows (at least one required)
- [ ] Functional testing

Fixes #

### DIFF
--- a/agasc/supplement/magnitudes/update_mag_supplement.py
+++ b/agasc/supplement/magnitudes/update_mag_supplement.py
@@ -335,6 +335,8 @@ def write_obs_status_yaml(obs_stats=None, fails=(), filename=None):
                 'comments': fail['msg']
             })
     if len(obs) == 0:
+        if filename and filename.exists():
+            filename.unlink()
         return
 
     yaml_template = """obs:


taldcroft:
Looks fine. The only quibble is that this assumes `filename` is a `Path`. That's fine, but the lack of a docstring makes this subtlety a bit hard to guess (esp. since `filename` implies a string instead of e.g. `filepath`).

BTW, I'm a fan of `Path.write_text()` to remove the boilerplate `with open(filename, 'w') as fh: fh.write(..)`.

javierggt:
yes, I actually went to the only place where this is called (I believe) to check it was a Path. I can change the docstring.

One could also program defensively and always do `filename = Path(filename)`, which is normally my tendency, but this adds a useless statement :)

taldcroft (approved):

taldcroft:
> program defensively and always do filename = Path(filename)

👍 on that. I like the idiom of delegating to `Path` (or other generically-useful classes like `CxoTime`) on input to allow the class initializer to validate the input and then get a nice predictable object.

taldcroft:
This then leads to using type definitions like `path-like` or `time-like` in the docs. See e.g. https://docs.astropy.org/en/latest/glossary.html?highlight=time-like

PR #125, opened 2021-08-05
Enhanced output obs_status to include maxmag

### LINKED ISSUE #116: Consider pre-populating more of the obs_status.yml file
Most (probably 90% or more) of the dispositions will be `Never acquired. Set mag to MAXMAG`. Finding the MAXMAG manually is a bit of a pain and error prone. (In fact I almost made a mistake in doing it this week because I was fooled by a color=1.5 star that I assumed was the bad one).

Since it is easier to delete or slightly modify YAML content then it is to compose it by hand, I propose that the auto-generated obs status file look like below. The comments
```
obs:
  - mp_starcat_time: 2021:154:16:23:25.445
    obsid: 46314
    status: 1
    agasc_id: [813964024]
    comments: Never acquired. Set mag to MAXMAG  
    ¡¡ CONFIRM or change !!

bad:
  813964024: 9  
  ¡¡ CONFIRM or change !!

mags:
  - agasc_id: 813964024
    mag_aca: 9.812
    mag_aca_err: 0.1
    ¡¡ CONFIRM or change !!
```
In essence make the whole thing correct for the most common disposition and then allow the reviewer to edit or accept.

The slightly funky `¡¡ CONFIRM or change !!` is negotiable, but what this does is makes the file be invalid YAML so it cannot be accidentally ingested without the reviewer specifically accepting each entry.

javierggt:
## Description

Following the suggestion in issue #116, this PR enhances the output obs_status.yml file to include some standard information that is often needed anyway.

The main differences with the suggestion in #116:
- there is no string invalidating the yaml so it cannot be ingested by mistake.
- The default "bad" value is 0, which is not a value in the list.
- The default magnitude is the MAXMAG value for the observation.
- the comments are not set to 'Never acquired. Set mag to MAXMAG'.

## Testing

- [x] Functional testing. I did the following using both the `master` version and the one in this PR:
     ```
     mkdir test-pr125
     cd test-pr125
     echo 5115672 > agasc_ids
     agasc-update-magnitudes --agasc-id-file agasc_ids
     ``` 
     with master I get the following yaml:
     ```
     obs:
       - mp_starcat_time: 2013:323:16:25:05.901
         obsid: 53222
         status: 1
         agasc_id: [5115672]
         comments: Suspect observation
     ```
     and with this branch I get:
     ```
     bad:
       5115672: 0
     mags:
     - agasc_id: 5115672
       mag_aca: 12.297
       mag_aca_err: 0.1
     obs:
     - agasc_id:
       - 5115672
       comments: Suspect observation
       mp_starcat_time: 2013:323:16:25:05.901
       obsid: 53222
       status: 1
     ```
     **Incidentally**... if you look into the current supplement, the magnitude is listed as 11.422, and the obs table says for this observation: "Faint star. Never acquired. Set mag to MAXMAG", but 11.422 is not the right value for MAXMAG for this specific observation. It is the MAXMAG in OBSID 8619, which is not currently flagged as suspect. The current version gives a magnitude of 12.623448 using OBSID 8619.

Fixes #116

### DIFF
--- a/agasc/supplement/magnitudes/update_mag_supplement.py
+++ b/agasc/supplement/magnitudes/update_mag_supplement.py
@@ -348,7 +348,7 @@ def write_obs_status_yaml(obs_stats=None, fails=(), filename=None):
         else:
             agasc_ids += [(agasc_id, -1) for agasc_id in obs['agasc_id']]
 
-        agasc_ids = dict(sorted(agasc_ids))
+    agasc_ids = dict(sorted(agasc_ids))
 
     yaml_template = """bad:
   {%- for agasc_id, maxmag in agasc_ids.items() %}
@@ -372,10 +372,11 @@ obs:
   {%- endfor %}
   """
     tpl = jinja2.Template(yaml_template)
+    result = tpl.render(observations=obs, agasc_ids=agasc_ids)
     if filename:
         with open(filename, 'w') as fh:
-            fh.write(tpl.render(observations=obs, agasc_ids=agasc_ids))
-    return tpl.render(observations=obs)
+            fh.write(result)
+    return result
 
 
 def do(start,
--- a/agasc/supplement/magnitudes/update_mag_supplement.py
+++ b/agasc/supplement/magnitudes/update_mag_supplement.py
@@ -14,6 +14,7 @@ import numpy as np
 from astropy import table
 from astropy import time, units as u
 
+from mica.starcheck import get_starcheck_catalog
 from agasc.supplement.magnitudes import star_obs_catalogs, mag_estimate, mag_estimate_report as msr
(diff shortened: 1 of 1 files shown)


javierggt:
For reference, the place where the "comments" are set, in the case of suspect observations, is [here](https://github.com/sot/agasc/blob/3b7010019ad6a3d323b0908da9bf03f2e2250cf0/agasc/supplement/magnitudes/mag_estimate.py#L1032). This is a "catch-all" case that might include cases where the star was tracked poorly. The question is whether we want to change the logic here to provide a better message.

I would vote **no**

taldcroft (approved):
LGTM. For kicks I did read in the new YAML on-the-side and it parsed to the expected data structure.

PR #126, opened 2021-08-05
Make sure observations with status=0 are not flagged suspect

### LINKED ISSUE #121: Observations with status=0 still flagged as suspect
The following AGASC IDs have observations with `status=0` in the `obs` table of the supplement. These observations apparently are still labeled as suspect and not included in magnitude estimate: 35263288,  46794712, 107743264, 118096888, 122159368, 157951008, 168038608, 190716896, 259801408, 262408104, 500706352, 612763008, 616432288, 651037128, 676336280, 904010616, 974925328, 1158028568

javierggt:
make sure observations with OK-status in status-override are OK regardless of what the algorithm says

## Description

Currently, observations with `status != 0` in the `obs` table of the supplement are excluded from magnitude estimates, but it turns out that observations with `status == 0` are not included (Issue #121). The expected behavior is that if `status == 0`, then they should be marked `OK` even if it fails the standard cuts. **This PR fixes that.**

The fix is simple. It mostly follows what is done for excluded observations: there is a list of _excluded_ observations, which are observations in the `obs` table with `status != 0`, so I added a corresponding list of _included_ observations. The boolean OK flag considers both.

Most of the changes are for testing:
- a test function
- a test data HDF5 file
- a function to re-create the test data file

## Testing

- [ ] ~~Functional testing~~

Fixes #121


taldcroft (approved):
LGTM

jeanconn (approved):
LGTMT

PR #129, opened 2021-12-29
Add comments to obs_status.yml with process help

taldcroft:
## Description

This PR puts much of the necessary information for supplement disposition and promotion into the RC `obs_status.yml` file.

## Testing

- [n/a] Passes unit tests on MacOS, linux, Windows (at least one required)
- [x] Functional testing

### Functional testing

From @javierggt:

I did the following test:
```
mkdir test
cd test
echo 5115672 > agasc_ids
agasc-update-magnitudes --agasc-id-file agasc_ids
```
and checked that the obs_status file was there and with the expected output. Then I edited the file, amended the supplement in the current directory, and checked that it was ok:
```
agasc-update-supplement --output-dir `pwd` --obs-status-file obs_status.yml 
```

### DIFF
--- a/agasc/supplement/magnitudes/update_mag_supplement.py
+++ b/agasc/supplement/magnitudes/update_mag_supplement.py
@@ -357,6 +357,8 @@ def write_obs_status_yaml(obs_stats=None, fails=(), filename=None):
 #
 # After reviewing and updating this file, run as `aca`:
 #   agasc-supplement-tasks disposition
+# Check the diffs at:
+#   https://cxc.cfa.harvard.edu/mta/ASPECT/agasc/supplement/agasc_supplement_diff.html
 # After reviewing the post-disposition email with diffs, run as `aca`:
 #   agasc-supplement-tasks schedule-promotion
 #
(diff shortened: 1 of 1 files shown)


javierggt:
Looks ok to me. Since we are adding a detailed description of steps to follow, we might as well add a link to [the diff](https://cxc.cfa.harvard.edu/mta/ASPECT/agasc/supplement/agasc_supplement_diff.html) that is generated after applying the dispositions.

As a test, one can run something like what is run in the workflow, which should produce an obs_status file, and then apply that to make sure it runs ok. I can do that. I do not foresse a problem, given that all changes are in comments.

taldcroft:
@javierggt - I added a link to the diffs per your suggestion. If you can do some functional test that would be great!

javierggt:
I did the following test:
```
mkdir test
cd test
echo 5115672 > agasc_ids
agasc-update-magnitudes --agasc-id-file agasc_ids
```
and checked that the obs_status file was there and with the expected output. Then I amended the supplement in the current directory and checked that it was ok:
```
agasc-update-supplement --output-dir `pwd` --obs-status-file obs_status.yml 
```

javierggt (approved):

PR #132, opened 2022-04-04
Do not count non-npm times when calculating f_ok

### LINKED ISSUE #131: Supplement processing does not handle SRDC's exactly right.
Take obsid 45684 as an example. There is an SRDC which causes all stars to be dropped around 2022:056:19:41:24. This lasts until 19:48:21 when there is a search command and stars return, so we have a ~7 (420 sec) minute gap. The observation kalman interval is from 2022:056:19:14:58.321 to 2022:056:20:07:26.097, so the observation is 3147 seconds long.

The [supplement report](https://cxc.cfa.harvard.edu/mta/ASPECT/agasc/supplement_reports/stars/086/869141880) shows 1071 samples with N_ok=649, where the gap is clearly from the SRDC missing data. These are 8x8 images so each image takes 4.1 sec. But 4.1 sec/readout * 1071 readouts is much more than 3147 sec.

The issue comes from counting the time interval of the SRDC in the denominator when calculating f_ok. According to the numbers above, the readouts occur every 4.1 seconds during the observation (about 2727 seconds) and every 1.025 seconds during the SRDC (about 420 seconds), which totals about 1075 samples.

f_ok should only include samples in the observation (not the SRDC), so this time interval should not be in the denominator.

javierggt:
## Description

This PR changes the calculation of f_ok in the magnitudes supplement so the SRDC times are not counted. While doing this, two other small changes were added:
- replace np.sum by np.count_nonzero where np.sum was applied on boolean arrays
- add f_ok_5 and f_ok_3 as the *explicit* ok flags within 3 and 5 arcsec.

Fixes #131

NOTE: also updated test data according  to column renames:
```
from astropy import table
telem = table.Table.read('agasc/tests/data/mag-stats.h5', path='obs_status/telem')
telem.rename_columns(['ok'], ['mag_est_ok'])
telem.write('agasc/tests/data/mag-stats.h5', path='obs_status/telem', append=True, overwrite=True, serialize_meta=True)
```

## Interface impacts

This adds `f_ok_3` and `f_ok_5` columns to the output fits files (not the supplement).

### Unit tests
- [x] Mac OSX

- [ ] [PLATFORM]:

#### Test 3

I looked at the differences between this PR and master for all observations in 2022. I am not worried about the differences. As expected, this version gives slightly higher values of `f_ok`, which causes a bit more observations to be considered. In terms of magnitudes, not much changes.

I made [this notebook](https://nbviewer.org/urls/cxc.cfa.harvard.edu/mta/ASPECT/ipynb/agasc/test-pr-132/agasc_pr-132.ipynb) which might not be very friendly.

#### Test 2
I set PYTHONPATH to my local directory, then I ran the following with the `master` and `srdc` branches:
```
agasc-update-magnitudes --agasc-id-file agasc_ids --report
```
where the `agasc_ids` file was just this:
```
869141880
38143784
```
The resulting supplement update reports are in https://cxc.cfa.harvard.edu/mta/ASPECT/jgonzalez/agasc/test-pr-132

I also compared the resulting `agasc_stats` and `obs_stats` as text and the differences make sense. The files are in the same directory as the reports.

#### Test 1:

I ran the following add-hoc test with the AGASC ID from #131:
```python
(ska3-flight) javierg SAO $ cat test_agasc_srdc.py
import pickle
import agasc
from agasc.supplement.magnitudes import mag_estimate as me
from agasc.supplement.magnitudes import star_obs_catalogs
import pprint

print(f'Using AGASC {agasc.__version__}')

stats = me.get_agasc_id_stats(869141880)
obs = star_obs_catalogs.STARS_OBS[star_obs_catalogs.STARS_OBS['agasc_id'] == 869141880][0]
telem = me.get_telemetry(obs)
obs_stats = me.calc_obs_stats(telem)

with open(f'agasc_stats-{agasc.__version__}.txt', 'w') as fh:
    pprint.pprint(stats, stream=fh)

with open(f'obs_stats-{agasc.__version__}.txt', 'w') as fh:
    pprint.pprint(obs_stats, stream=fh)
```

I ran it like this and got the expected result:
```bash
(ska3-flight) javierg SAO $ python test_agasc_srdc.py
Using AGASC 4.11.4
(ska3-flight) javierg SAO $ export PYTHONPATH=`pwd`/git/agasc
(ska3-flight) javierg SAO $ python test_agasc_srdc.py
Using AGASC 4.11.6.dev7+g7898a58
(ska3-flight) javierg SAO $ diff -U 0 agasc_stats-4.11.4.txt agasc_stats-4.11.6.dev7+g7898a58.txt 
--- agasc_stats-4.11.4.txt	2022-04-04 16:37:05.000000000 -0400
+++ agasc_stats-4.11.6.dev7+g7898a58.txt	2022-04-04 16:37:32.000000000 -0400
@@ -5 +5,3 @@
-  'f_ok': 0.6059757236227824,
+  'f_ok': 1.0,
+  'f_ok_3': 1.0,
+  'f_ok_5': 1.0,
@@ -20 +22 @@
-  'n': 1071,
+  'n': 649,
(ska3-flight) javierg SAO $ diff -U 0 obs_stats-4.11.4.txt obs_stats-4.11.6.dev7+g7898a58.txt 
--- obs_stats-4.11.4.txt	2022-04-04 16:37:05.000000000 -0400
+++ obs_stats-4.11.6.dev7+g7898a58.txt	2022-04-04 16:37:32.000000000 -0400
@@ -12 +12,3 @@
- 'f_ok': 0.6059757236227824,
+ 'f_ok': 1.0,
+ 'f_ok_3': 1.0,
+ 'f_ok_5': 1.0,

```


agasc/supplement/magnitudes/mag_estimate.py
```diff
@@ -722,18 +722,19 @@ def calc_obs_stats(telem):
 
     kalman = (telem['AOACASEQ'] == 'KALM') & (telem['AOPCADMD'] == 'NPNT')
     track = (telem['AOACIIR'] == 'OK') & (telem['AOACFCT'] == 'TRAK')
```
jeanconn, line 724: I have forgotten our process so maybe this should have a comment about why saturated pixel is not included in the filter?

agasc/supplement/magnitudes/mag_estimate.py
```diff
@@ -722,18 +722,19 @@ def calc_obs_stats(telem):
...
+    dr3 = (telem['dy'] < 3) | (telem['dz'] < 3)
+    dr5 = (telem['dy'] < 5) | (telem['dz'] < 5)
 
-    f_kalman = np.sum(kalman) / len(kalman)
-    n_kalman = np.sum(kalman)
-    f_track = np.sum(kalman & track) / n_kalman if n_kalman else 0
-    n_track = np.sum(kalman & track)
-    f_3 = (np.sum(kalman & track & dr3) / n_track) if n_track else 0
-    f_5 = (np.sum(kalman & track & dr5) / n_track) if n_track else 0
+    f_kalman = np.count_nonzero(kalman) / len(kalman)
```
jeanconn, line 728: Not sure if this f_kalman is useful.

agasc/supplement/magnitudes/mag_estimate.py
javierggt: It would be a good thing to remember before we try to get a Gaia-based catalog though.

javierggt:
About whether to include saturated pixels in f_ok...

In the development of this code, there was the tacit assumption the the `OK` flag referred to whether that specific sample is "OK" for the purpose of estimating the magnitude. An *OK* sample gets included in the magnitude estimation, whereas a *not-OK* one does not. In other words: the *OK fraction* `f_ok` refers to the fraction of time one has an image that is used to estimate the magnitude.

It seems like now you want to interpret that `f_ok` in a more general way, which was not the intent, and I would prefer not to change it. If you want, I can add another flag with the definition you are talking about now.

Excluding times with SP in the `f_ok` calculation will reduce the value of `f_ok` and more stars will be marked as suspect. Some stars with many saturated pixels would be excluded. This was my original definition of `f_ok` and we scratched it because we would prefer to have some magnitude information for these bright stars rather than not.

A separate issue is whether we should improve the magnitude determination to properly handle SP.

I have modified `f_ok_3` and `f_ok_5` in how the angular residual is checked, but I don't think we should change more than that.

agasc/supplement/magnitudes/mag_estimate.py
```diff
@@ -722,18 +722,19 @@ def calc_obs_stats(telem):
 
     kalman = (telem['AOACASEQ'] == 'KALM') & (telem['AOPCADMD'] == 'NPNT')
     track = (telem['AOACIIR'] == 'OK') & (telem['AOACFCT'] == 'TRAK')
-    dr3 = (telem['dr'] < 3)
-    dr5 = (telem['dr'] < 5)
+    dr3 = (telem['dy'] < 3) | (telem['dz'] < 3)
```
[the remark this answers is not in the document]
javierggt: I would have never guessed that, but if it helps I will change it. We would be changing only `dr5` to use a square metric.

Still... strictly speaking these flags do not have to correspond to whatever the OBC was thinking the residual was. I believe that adding the square cut exactly is not going to make it agree with the OBC expectation more because the centroid positions might be different.

In any case, if you *really* want the square cut, I will change it.

agasc/supplement/magnitudes/mag_estimate.py
javierggt: I wrote a global comment responding to this.

taldcroft:
> It seems like now you want to interpret that f_ok in a more general way, which was not the intent, and I would prefer not to change it. If you want, I can add another flag with the definition you are talking about now.

My intent is not to change *anything* about magnitude estimation. I think that the way the code is written now is correct, and in particular we are agreed that SP should be ignored for mag estimation. (And that's my understanding of the current code).

> Excluding times with SP in the f_ok calculation will reduce the value of f_ok and more stars will be marked as suspect. Some stars with many saturated pixels would be excluded. This was my original definition of f_ok and we scratched it because we would prefer to have some magnitude information for these bright stars rather than not.

Agreed.

> A separate issue is whether we should improve the magnitude determination to properly handle SP.

I don't think so.

> I have modified f_ok_3 and f_ok_5 in how the angular residual is checked, but I don't think we should change more than that.

I think we're on the same page but see #134.

taldcroft:
I wrote up a glossary of definitions in language that makes sense for me. Even if the audience is just future-me, can you make something with these definitions easily accessible from the star detail page? I find the Table hover-overs to be not that helpful in practice since they are not formatted (e.g. it is rendered as plain text not HTML, and I always forget that you need to leave the cursor over the column name for a while and find myself getting frustrated moving the cursor around trying to the tooltip to appear). So I'd rather have something that says "Definitions" with a collapse/expand element which reveals this text.

It would be worth auditing the code to make sure the code matches my definitions. (I've already noted discrepancies in my code review).
```
sample
  One sample of ACA image telemetry. This could be a real image readout or it
  could be a null 4x4 image when the ACA is not tracking. Be aware that the
  sample period is faster for null 4x4 images (1.025 sec) than for typical 6x6
  or 8x8 tracked images (2.05 or 4.1 sec respectively).

Kalman samples
  Subset of samples when ACA could be tracking stars (AOPCADMD == NPNT &
  AOACASEQ == KALM)

dr3
  Subset of tracked Kalman samples with radial centroid residual < 3 arcsec.
  This corresponds to "high quality" readouts. This effectively includes the
  track subset (AOACFCT == TRAK) because readouts with no TRAK are assigned an
  infinite centroid residual.

dbox5
  Subset of tracked Kalman samples with centroid residual within a 5 arcsec box.
  This corresponds to spatial filter used by the OBC for inclusion in Kalman
  filter. This effectively includes the track subset (AOACFCT == TRAK) because
  readouts with no TRAK are assigned an infinite centroid residual.

track
  Subset of Kalman samples where the image is being tracked (AOACFCT == TRAK)

sat_pix
  Subset of Kalman samples with saturated pixel flag OK (AOACISP == OK)

ion_rad
  Subset of Kalman samples with ionizing radiation flag OK (AOACIIR == OK)

n_total
  Total number of sample regardless of OBC PCAD status

n
  Synonym for n_total

n_kalman
  Number of Kalman samples

n_dr3
  Number of dr3 samples.

n_dbox5
  Number of dr5 samples.

n_track
  Number of track samples.

n_ok_3
  Number of (track & sat_pix & ion_rad & dr3) samples

n_ok_5
  Number of (track & sat_pix & ion_rad & dbox5) samples

f_3
  Number of (track & dr3) samples / n_track

f_5
  Number of (track & dbox5) samples / n_track

f_dr3
  n_dr3 / n_kalman

f_dbox5
  n_dbox5 / n_kalman

f_ok
  n_ok_3 / n_kalman. This is a measure of the fraction of time during an
  observation that the Kalman filter is getting high-quality star centroids.

f_ok_3
  n_ok_3 / n_kalman. Same as f_ok.

f_ok_5
  n_ok_5 / n_kalman. This is a measure of the fraction of time during an
  observation that the Kalman filter is getting any star centroid at all. This
  includes measurements out to 5 arcsec box halfwidth, so potentially 7 arcsec
  radial offset.
 ```

javierggt:
@taldcroft, I made edits according to all comments you made before, then squashed commits to make history and diffs a bit easier to navigate.

In the first commit, I added your list of definitions ([this commit](https://github.com/sot/agasc/pull/132/commits/71a267687ed695277ba9e06cd3767652f62f60ec)) and the following commit are changes I made to this list ([this commit](https://github.com/sot/agasc/pull/132/commits/b285b14f1dd6fbee26c7e9d77aab8b2a67bf2282)). In that way we can first agree on what means what. If you agree with the edited list, then you can proceed to the [next commit](https://github.com/sot/agasc/pull/132/commits/b795f038f31852d302314183d2c480cd2492af8e), which is related to renaming variables and such (most things not originally in this PR).

taldcroft (approved):
:tada: thanks @javierggt this will definitely make the supplement products even more useful!

PR #133, opened 2022-04-04
Fix matplotlib ticks warning

javierggt:
## Description

Change code to avoid the following matplotlib warning:
```
FixedFormatter should only be used together with FixedLocator
```

I also made the following changes:
- added a line to `.gitignore`
- added --all-stars option to agasc-magnitudes-report (if not given, the report includes only suspect stars)

## Interface impacts
None

### Unit tests

- [x] Mac

Independent check of unit tests by @taldcroft 
- [x] Mac

### Functional tests

For testing, I added an option to the `agasc-magnitudes-report` (included in this PR) which causes the report to include all stars, not just the ones with suspect observations. I ran that script in both the `master` and `fix-warning` branches with the folllowing bash commands and looked at differences in the output:
```bash
git co fix-warning
cp agasc/scripts/mag_estimate_report.py .
python mag_estimate_report.py --start 2022:085:20:25:00 --output-dir fix-warning --all-stars
git co master
python mag_estimate_report.py --start 2022:085:20:25:00 --output-dir master --all-stars
for f in `find master -type f`; do echo ${f/master/fix-warning}; diff -u $f ${f/master/fix-warning}; done;
```

I could verify that:
- in `master` I got the warning whereas in `fix-warning` I did not.
- The output is the same.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ _*
 *~
 supplement_reports
 test
+.idea
--- a/agasc/scripts/mag_estimate_report.py
+++ b/agasc/scripts/mag_estimate_report.py
@@ -36,6 +36,9 @@ def get_parser():
     parser.add_argument('--weekly-report',
                         help="Add links to navigate weekly reports.",
                         action='store_true', default=False)
+    parser.add_argument('--all-stars',
+                        help="Include all stars in the report, not just suspect.",
+                        action='store_true', default=False)
     return parser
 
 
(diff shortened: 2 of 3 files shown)


jeanconn (changes requested):
Should really call out the all-stars option in the description, not just the testing (even if it is testing-focused).

taldcroft (approved):
Looks good. The comment from Jean about removing that leftover comment should be addressed and then this can be merged.

PR #134, opened 2022-04-05
Untested change for discussion

taldcroft:
## Description

This is a PR on the `srdc` branch that reflects how I think things should be defined, and also a level of code comments that allows me to understand the outputs.

I was going to try to describe all this but then found working on the code would be easier.

cc: @javierggt

### DIFF
--- a/agasc/supplement/magnitudes/mag_estimate.py
+++ b/agasc/supplement/magnitudes/mag_estimate.py
@@ -720,22 +720,47 @@ def calc_obs_stats(telem):
     """
     times = telem['times']
 
-    kalman = (telem['AOACASEQ'] == 'KALM') & (telem['AOPCADMD'] == 'NPNT')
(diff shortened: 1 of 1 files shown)


agasc/supplement/magnitudes/mag_estimate.py
```diff
@@ -720,22 +720,47 @@ def calc_obs_stats(telem):
     """
     times = telem['times']
 
-    kalman = (telem['AOACASEQ'] == 'KALM') & (telem['AOPCADMD'] == 'NPNT')
-    track = (telem['AOACIIR'] == 'OK') & (telem['AOACFCT'] == 'TRAK')
-    dr3 = (telem['dy'] < 3) | (telem['dz'] < 3)
-    dr5 = (telem['dy'] < 5) | (telem['dz'] < 5)
```
taldcroft, line 726: Note that I think this needs `np.abs()` and `&`, see below.

PR #137, opened 2022-05-02
Write JSON when creating star reports.

javierggt:
## Description

With the changes in this PR, whenever a star report is created, a JSON file is saved that contains the relevant obs-stats and agasc-stats (the input to the template). This is not the only possible way to accomplish what we want. We could also just use the fits files in `$SKA/data/agasc`.

To accomplish this, I added a JSON encoder to convert the tables to json.

## Interface impacts
None

### Unit tests

**I added unit tests for the new methods.**
- [ ] No unit tests
- [x] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests

I ran the following: 
```
python -m agasc.scripts.mag_estimate_report --start 2022:080:02:00 --stop 2022:81:00:00 --all-stars
```
and verified that the json files were created and readable. The files (with `indent=2`, which is not exactly how they are produced) are:
- [top-level](https://cxc.cfa.harvard.edu/mta/ASPECT/jgonzalez/agasc/pr-137/index.json)
- [star](https://cxc.cfa.harvard.edu/mta/ASPECT/jgonzalez/agasc/pr-137/data.json)

### DIFF
--- a/agasc/supplement/magnitudes/mag_estimate_report.py
+++ b/agasc/supplement/magnitudes/mag_estimate_report.py
@@ -53,16 +53,11 @@ class TableEncoder(json.JSONEncoder):
     """
     def default(self, obj):
         if isinstance(obj, table.Table):
-            return {k: obj[k] for k in obj.colnames}
-        if isinstance(obj, table.MaskedColumn):
-            fv = obj.get_fill_value()
-            return [(fv if mask else val) for mask, val in zip(obj.mask, obj)]
-        if isinstance(obj, table.Column):
-            return [val for val in obj]
-        if isinstance(obj, table.Row):
-            return obj.values()
-        if isinstance(obj, np.ma.core.MaskedConstant):
-            return obj.view().fill_value
+            return {name: val.tolist() for name, val in obj.columns.items()}
+        if isinstance(obj, CxoTime):
+            return obj.isot
+        if isinstance(obj, np.ma.core.MaskedArray):
+            return {'columns': obj.dtype.names, 'data': obj.tolist()}
         if np.isscalar(obj):
             return int(obj)
         if np.isreal(obj):
@@ -154,10 +149,11 @@ class MagEstimateReport:
                 {
                     'agasc_stats': agasc_stat,
                     'obs_stats': obs_stat,
-                    'static_dir': static_dir
+                    'static_dir': static_dir,
+                    'glossary': GLOSSARY,
                 },
                 json_out,
-                cls=TableEncoder
+                cls=TableEncoder,
             )
         return directory / 'index.html'
 
@@ -284,6 +280,24 @@ class MagEstimateReport:
                                           tooltips=tooltips,
                                           static_dir=static_dir,
                                           glossary=GLOSSARY))
+
+        json_filename = filename.replace('.html', '.json')
+        if json_filename == filename:
+            json_filename = filename + '.json'
+        with open(self.directory / json_filename, 'w') as json_out:
+            json.dump(
+                {
+                    'info': info,
+                    'sections': sections,
+                    'failures': fails,
+                    'star_reports': star_reports,
+                    'tooltips': tooltips,
+                    'static_dir': static_dir,
+                    'glossary': GLOSSARY,
+                },
+                json_out,
+                cls=TableEncoder,
+            )
 
     def plot_agasc_id_single(self, agasc_id, obsid=None,
                              telem=None,
--- a/agasc/supplement/magnitudes/mag_estimate_report.py
+++ b/agasc/supplement/magnitudes/mag_estimate_report.py
@@ -17,7 +17,6 @@ from tqdm import tqdm
 from astropy import table
 from cxotime import CxoTime
 
-from agasc.supplement import utils
 from agasc.supplement.magnitudes import mag_estimate
 
 
@@ -27,6 +26,48 @@ JINJA2 = jinja2.Environment(
 )
 
 logger = logging.getLogger('agasc.supplement')
+
+
+class TableEncoder(json.JSONEncoder):
+    """
+    Utility class to encode tables as json.
+
+    Example::
+
+        >>> import json
+        >>> from astropy.table import Table, Column, MaskedColumn
+        >>> a = MaskedColumn([1, 2], name='a', mask=[False, True], dtype='i4')
+        >>> b = Column([3, 4], name='b', dtype='i8')
+        >>> t = Table([a, b])
+        >>> print(json.dumps(t, cls=TableEncoder, indent=2))
+        {
+          "a": [
+            1,
+            999999
+          ],
+          "b": [
+            3,
+            4
+          ]
+        }
+    """
+    def default(self, obj):
+        if isinstance(obj, table.Table):
+            return {k: obj[k] for k in obj.colnames}
+        if isinstance(obj, table.MaskedColumn):
+            fv = obj.get_fill_value()
+            return [(fv if mask else val) for mask, val in zip(obj.mask, obj)]
+        if isinstance(obj, table.Column):
+            return [val for val in obj]
+        if isinstance(obj, table.Row):
+            return obj.values()
+        if isinstance(obj, np.ma.core.MaskedConstant):
+            return obj.view().fill_value
+        if np.isscalar(obj):
+            return int(obj)
+        if np.isreal(obj):
+            return float(obj)
+        return json.JSONEncoder.default(self, obj)
 
 
 class MagEstimateReport:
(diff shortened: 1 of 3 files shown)


agasc/supplement/magnitudes/mag_estimate_report.py
```diff
@@ -107,6 +109,16 @@ def single_star_html(self, agasc_id, directory,
                                            obs_stats=o.as_array(),
                                            static_dir=static_dir,
                                            glossary=GLOSSARY))
+        with open(directory / 'data.json', 'w') as json_out:
+            json.dump(
+                {
+                    'agasc_stats': self.agasc_stats[self.agasc_stats['agasc_id'] == agasc_id],
+                    'obs_stats': o,
```
taldcroft, line 116: This crosses the threshold of no banned single-letter variable names like `o` in new code. You can use VS code to rename this to `obj_stat` and the code will suddenly be more readable to me.

taldcroft (commented):
Awesome, nice! I didn't try this out yet but at first glance it looks good. 

I think it would be worth putting the nice new `TableEncoder` into a more general location, maybe in `ska_helpers.utils`. I had not quite been expecting tables of data there, but I hadn't thought about it too hard. Can you put an example JSON file somewhere to have a look at the content?

agasc/supplement/magnitudes/mag_estimate_report.py
javierggt: There is an example output in the docstring, but if you want to see an actual file related to this PR, I can show you. These files are not indented so they are not so readable. I made a copy with indentation here: https://cxc.cfa.harvard.edu/mta/ASPECT/jgonzalez/agasc/example_data.json

taldcroft:
I don't understand why some of the columns in `agasc_stats` are masked. Most obviously `agasc_id`, where I cannot see having an entry in the table without a valid `agasc_id`.  Likewise most or all of the counts `n_<something>` should never be missing. I can see how something like `t_mean_dr3_not` might not be defined, assuming that references the mean mag for samples that don't satisfy `dr3`.  (What is the `t_` prefix in general?).

 In any case, maybe it is reasonable to write out `agasc_stats` as a dict and use `None` for anything that is masked.

taldcroft:
One thing to maybe consider is writing a JSON serializer that is bit more standards-driven. E.g. here is one of the pandas JSON formats:
```
In [32]: t = simple_table(masked=True)

In [33]: tp = t.to_pandas()

In [34]: tp.to_json('junk.json', orient='table', index=False)

In [35]: cat junk.json
{"schema":
{"fields":[{"name":"a","type":"integer"},{"name":"b","type":"number"},
{"name":"c","type":"string"}],"pandas_version":"0.20.0"},
"data":[{"a":null,"b":1.0,"c":"c"},{"a":2,"b":2.0,"c":null},{"a":3,"b":null,"c":"e"}]}
In [36]: t
Out[36]: 
<Table masked=True length=3>
  a      b     c  
int64 float64 str1
----- ------- ----
   --     1.0    c
    2     2.0   --
    3      --    e
```
This is basically an implementation of http://dataprotocols.org/json-table-schema/.

taldcroft:
I was thinking to just use Pandas completely, but curiously if you try to read that same file back with pandas it fails with `IntCastingNaNError: Cannot convert non-finite values (NA or inf) to integer`. Huh.

javierggt:
I see the value in a more general table encoder, but that was not my intent with this PR. I started this now because I wanted to start reprocessing, and wanted to add the JSON feature, and I wanted it now. There are a bunch of considerations that go into a more general encoder (formats, fill values, meta-data... you know better than me).

One problem with using `None` as the fill value is that it would not work directly with the template, because the template includes formats. So here the question is "what do we want this for? to save general table info or to save the input to the template?".

I am still not convinced this is the best way. To begin with, we could just use the fits files, without any json file. Zero lines of code in agasc, twenty on web-kadi-apps. Smaller than this PR. If we *really* wanted a json file, I am adding a simpler encoder that does the job for this use case and requires no decoder.

Even after any file-related changes, there is still the issue of plugging the template from agasc into the kadi template. The agasc template uses foundation, and the agasc template uses bootstrap. Is it going to work? I am betting there will be issues.

The natural answer is to ditch the permanent agasc star pages and just move them into web-kadi-apps, so there is one template (with whatever bootstrap/foundation issues ironed out).

In summary I suggest:
- we move the discussion  of a generic encoder to ska_helpers.
- we tag current master and start reprocessing.
- close this PR.
- we see about moving the single-star reports to web-kadi-apps (and perhaps the weekly reports too).

taldcroft:
I'm confused by your concern about foundation and templates. The idea (and what your code actually does right now), is serialize the inputs to this code that currently generates the static HTML within `agasc`.
```
        with open(directory / 'index.html', 'w') as out:
            out.write(star_template.render(agasc_stats=s,
                                           obs_stats=o.as_array(),
                                           static_dir=static_dir,
                                           glossary=GLOSSARY))
```
So within `web-kadi-apps` it seems trivial and robust to just de-serialize those data values for a particular star ID and run `star_template.render(...)`, taking `star_template` from either `agasc` or from an app in `web-kadi-apps`. Basically I'm missing what you think might be a problem.

taldcroft:
The reason I am against just reading from the FITS file is that they are big (140 Mb in total) and you really don't want to read for each web page access just to get a handful of numbers.

javierggt:
> So within web-kadi-apps it seems trivial and robust to just de-serialize those data values for a particular star ID and run star_template.render(...), taking star_template from either agasc or from an app in web-kadi-apps. Basically I'm missing what you think might be a problem.

What you describe is the data part of the process. Now the rendering...

The current blueprint renders a page using the kadi template. This template includes and iframe which loads the agasc webpage (the agasc template statically rendered). Both of these templates are complete web pages, with header and body. Each uses a different framework for layout and style.

If you just do `star_template.render(...)`, you are not going to get the kadi page, with the dialog to get another star, as it is now. If that is what you want, then we do not a need json file either, because a redirect will suffice.

If you want the kadi look and feel, with the form to get a page for an agasc id, then you need to somehow embed the template from agasc into the kadi template. That is where the foundation/bootstrap issues might appear.

taldcroft:
Sorry, I was speaking a bit loosely and not thinking of literally using the current `star_template.render` which makes the entire web page. I was thinking of embedding the content into a standard web-kadi template that provides the styling and header/footer etc.

We explicitly want a new template to allow flexibility in the content and design. So I also rescind the comment about taking the template directly from `agasc`, that was wrong. The existing `agasc` template `star_report.html` would be ported into `web-kadi-apps` and then retired.

Those star report pages are really simple, just a few tables and an image. Probably if you cut out the bootstrap and just let the existing Foundation take over then it seems like all will be well?

javierggt:
Just in case it is not clear: I am working on this. I will refresh the review request when ready.

javierggt:
The code in this PR works and I have tested it after implementing some changes suggested by @taldcroft. Now the question is whether we want to merge them. I am not too convinced this actually is very useful. It does not hurt...

taldcroft:
@javierggt - if you don't see an immediate driver for having this I'd say to leave it open for now. After a few months we can revisit and either merge or close.

javierggt:
The driver is the same as before: generating a page in to display this data, but the reason why it might not be useful is that kadi has access to the file system, so it can get the data from the fits files on SKA/data. It is "nice" to have the json files handy, but it is not required.

taldcroft:
@javierggt - since you are the creator of this content and most likely the consumer, it is up to you. I looked at the code and I don't see any way this can break the core functionality of supplement processing, so I approve this if you want to merge.

taldcroft (approved):

agasc/supplement/magnitudes/mag_estimate_report.py
```diff
@@ -27,6 +28,48 @@
...
+        if isinstance(obj, table.Column):
+            return [val for val in obj]
+        if isinstance(obj, table.Row):
+            return obj.values()
+        if isinstance(obj, np.ma.core.MaskedConstant):
+            return obj.view().fill_value
+        if np.isscalar(obj):
+            return int(obj)
+        if np.isreal(obj):
+            return float(obj)
```
taldcroft, line 69: I think this can be replaced with the one-liner:
```suggestion
        if isinstance(obj, table.Table):
            return {name: val.tolist() for name, val in obj.columns.items()}
```
This puts `None` in the output for masked values, which is fine because JSON serializes that as `null`. That captures the intent of the missing value more than using the fill value which is not easy for downstream applications to handle (meaning to get back to the original masked column).

The only thing this version does not do is serialize a Table Row, so that would be separate. I'm not sure why that was included since it seems out of scope from the docstring "encode tables as json".

javierggt: I added the row conversion because I needed it. If anything, the docstring is outdated. I would have to go back and check where exactly a row was serialized.

javierggt: In the end, I followed this suggestion.

javierggt:
I rebase on master and added code to also save the top-level information to JSON.

taldcroft (approved):

PR #135, opened 2022-04-12
Use commands v2 for AGASC supplement update

javierggt:
## Description

This PR completely removes all uses of timelines and kadi commands v1.

A preliminary exploration of the differences in the two observation lists (from commands v1 and v2) can be found [here](https://nbviewer.org/urls/cxc.cfa.harvard.edu/mta/ASPECT/ipynb/agasc/agasc-pr-135.ipynb). We went over that notebook a while ago, and since then kadi commands v2 became the standard, so at this point we trust that the observations are properly defined.

to do:
- [x] `test_obs_status.py::test_override` fails.

## Interface impacts

The columns of `star_obs_catalogs.STARS_OBS` will change (see notebook).

Consequently the columns in `mag_stats_obsid.fits` file also change. This file can be updated during promotion by running the following script on it:
   ```
   from astropy import table;
   master = table.Table.read('mag_stats_obsid.fits');
   master.remove_column('timeline_id');
   master.write('mag_stats_obsid.fits', overwrite=True);
   ```

## Testing

Comprehensive functional test is described below.

### Unit tests

- [x] Mac

Independent check of unit tests by [REVIEWER NAME]
- [ ] [PLATFORM]:

### Functional tests

- [x] starting with an empty directory:
   ```
   python -m agasc.scripts.update_mag_supplement --start 2021:011 --stop 2021:012
   ```
- [x] starting with an empty directory. In this case, the `--multi-process` option failed in master, not in the branch.
   ```
   python -m agasc.scripts.update_mag_supplement --start 2022:001 --stop 2022:090 [--multi-process]
   ```
- [x] This time range includes the latest safe mode. This test failed both in master and the branch when given the `--report` option. I did not fix it because: a) it might be related to the safe mode, b) it is unrelated to this PR because it failed in both cases.
   ```
   python -m agasc.scripts.update_mag_supplement --start 2023:040 --stop 2023:050
   ```
- [x] The following tests a typical weekly update, where I had to change the `mag_stats_obsid.fits` file to remove the `timeline_id` column:
   ```
   cp $SKA/data/agasc/agasc_supplement.h5 $SKA/data/agasc/rc/*.fits .
   ```
   - only in the branch case, ran this script to remove the `timeline_id` column from `mag_stats_obsid.fits`:
   ```
   from astropy import table;
   master = table.Table.read('mag_stats_obsid.fits');
   master.remove_column('timeline_id');
   master.write('mag_stats_obsid.fits', overwrite=True);
   ```
   - run:
   ```
[...]

PR #140, opened 2022-10-22
Use kadi commands v1 for now

javierggt:
## Description

Unfortunately, the agasc supplement processing hit a snag because PR #135 has not been merged.

PR #135 is the "proper" way of doing things and takes full advantage of commands v2, but it is deep into the way observations are defined, so it might require more testing. I would like to have a fix for the weekly processing for a release this week.

The errors in the weekly run are due to cmds v2 being used, and these set the commands version to '1' in all top-level scripts.

## Interface impacts
None

### Unit tests
- [ ] Mac

- [ ] [PLATFORM]:

### Functional tests

No functional testing yet. I will just run each script for one OBSID.


jeanconn:
Alternatively, how about setting  KADI_COMMANDS_VERSION=1 in the crontab lines for now?

javierggt:
that does not work, as you might have seen in your email.

I also tried
```
KADI_COMMANDS_VERSION=1 && task_schedule3.pl -config agasc/task_schedule_update_supplement_rc.cfg
```

and got the same error. The only other option is to run the individual commands in the task schedule

jeanconn:
I have forgotten most of my unix, but I always do "env VAR=thing command"

jeanconn:
So "env" is probably not necessary anymore but, I think using "&&" to conditionally run the next part of the command does not pass the environment variable through to the environment used by the command
```
jeanconn-fido> KADI_COMMANDS_VERSION=1 && printenv | grep KADI
jeanconn-fido> KADI_COMMANDS_VERSION=1 printenv | grep KADI
KADI_COMMANDS_VERSION=1
```

jeanconn:
I'm fine with the code changes but figured I should hold review until functional testing noted?

taldcroft (approved):
Code and functional testing LGTM

jeanconn (approved):

PR #143, opened 2022-11-02
Use union=True when filtering bad times in MSID set

### LINKED ISSUE #141: operands could not be broadcast together in get_mag_from_img
[This week there were a few failures](https://cxc.cfa.harvard.edu/mta/ASPECT/agasc/supplement_reports/weekly/2022:297/index.html) in several OBSIDs. The underlying errors were like this:
```
ValueError: operands could not be broadcast together with shapes (20946,) (20947,) 
```

The one thing that changed right before this error was that commands-v2 was set to be the default, and I started running with the `KADI_COMMANDS_VERSION` environment variable set to `1`. The error can be reproduced using this script:
```
from agasc.supplement.magnitudes import mag_estimate, star_obs_catalogs as cat
cat.load()
obs = cat.STARS_OBS[(cat.STARS_OBS['agasc_id'] == 109584104) & (cat.STARS_OBS['obsid'] == 5606)][0]
mag_estimate.get_telemetry(obs)
```

javierggt:
This PR introduces a change to fix #141. That issue arised because there is an MSID set which had bad times that were different for different MSIDs. This behavior in MSIDset.filter_bad was introduced in Ska.engarchive version 4.57.0.

To fix it, we now use `fetch.MSIDset` instead of `fetch.Msidset` and directly call MSIDset.filter_bad with the `union=True` argument.

## Description

Fixes #141 

## Interface impacts
None

### Unit tests
- [x] Mac

- [ ] [PLATFORM]:

### Functional tests

The following script fails before this PR and succeeds after:

```
from agasc.supplement.magnitudes import mag_estimate, star_obs_catalogs as cat
cat.load()
obs = cat.STARS_OBS[(cat.STARS_OBS['agasc_id'] == 109584104) & (cat.STARS_OBS['obsid'] == 5606)][0]
mag_estimate.get_telemetry(obs)
```

### DIFF
--- a/agasc/supplement/magnitudes/mag_estimate.py
+++ b/agasc/supplement/magnitudes/mag_estimate.py
@@ -297,7 +297,9 @@ def get_telemetry(obs):
     names = ['AOACASEQ', 'AOPCADMD', 'CVCMJCTR', 'CVCMNCTR',
              f'AOACIIR{slot}', f'AOACISP{slot}', f'AOACMAG{slot}', f'AOACFCT{slot}',
              f'AOACZAN{slot}', f'AOACYAN{slot}'] + [f'AOATTQT{i}' for i in range(1, 5)]
-    msids = fetch.Msidset(names, start, stop)
+    # not using Msidset because that calls filter_bad with union=False
+    msids = fetch.MSIDset(names, start, stop)
+    msids.filter_bad(union=True)
     if len(slot_data) == 0:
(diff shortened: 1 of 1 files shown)


taldcroft (approved):
This looks good with the documented functional testing. 

It isn't clear to me if #142 is still needed with this fix, but anyway it should be independent of this PR.

PR #144, opened 2022-11-08
Reduce peak memory use by not reading full file to get RA DEC columns

jeanconn:
## Description
Reduce peak memory use by reading just the RA and DEC fields to get them (used for indexing into the agasc).

### Unit tests

- [x] Mac

Independent check of unit tests by @taldcroft:
- [x] Mac

### Functional tests

I looked at this before and after with memory profiler output and confirmed that the change means that, using the proseco agasc file as an example, only the 60Mb columns are read into memory instead of the file (which peaks more at ~600Mb).

### DIFF
--- a/agasc/agasc.py
+++ b/agasc/agasc.py
@@ -119,14 +119,11 @@ class RaDec(object):
         return self._dec
 
     def read_ra_dec(self):
-        # Read the file of RA and DEC values (sorted on DEC):
-        #  dec: DEC values
-        #  ra: RA values
+        # Read the RA and DEC values from the agasc
         with tables.open_file(self.agasc_file) as h5:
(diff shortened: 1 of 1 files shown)


jeanconn:
I considered updating the code to remove the explicit cache, and provide an option to not-cache the RA DECs altogether... but maybe it makes sense to limit scope creep.

taldcroft (approved):
Looks good!

taldcroft:
And even though I'm now fully convinced that returning from within a context block will run the `__exit__` method (https://stackoverflow.com/questions/9885217/in-python-if-i-return-inside-a-with-block-will-the-file-still-close), I still think it smells bad. (i.e. it has a popular a stack overflow answer so it is confusing).

PR #182, opened 2024-06-26
Update readme for 1p8

javierggt:
## Description

This PR makes minimal changes to readme (the one in github.io). I tried to leave the readme as unchanged as possible, but I actually believe it should be improved more. I think as it stands, the readme is hard to understand.

Still, these changes would bring it up to date.

questions:
- should we document here that slight changes to COLOR1 to avoid 0.7 and 1.5?
- What is the alpha in the `MAG_BAND` description?

### Unit tests

- [ ] No unit tests
- [ ] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests

No functional testing.

### DIFF
--- a/docs/catalog-description.rst
+++ b/docs/catalog-description.rst
@@ -297,7 +297,7 @@ following fields for each entry::
         For Tycho catalogs, this is the Tycho BT-VT color.
         Default value of -9999.0
 
-    2    COLOR2_ERR - short integer expressing the error in color2, iun
+    2    COLOR2_ERR - short integer expressing the error in color2, in
         units of 0.01mag.    Default value of -9999.
 
     1    C2_CATID - unsigned integer identifying the source of color2 and
--- a/docs/catalog-description.rst
+++ b/docs/catalog-description.rst
@@ -1,10 +1,10 @@
 ================================================
-THE AXAF Guide and Acquisition Star Catalog V1.7
+THE AXAF Guide and Acquisition Star Catalog V1.8
 ================================================
 
 An all-sky astrometric and photometric catalog
 prepared for the operation of Chandra
-(formerly AXAF). version 1.7 September 2018
+(formerly AXAF). version 1.8 July 2024
 
 Introduction
 ============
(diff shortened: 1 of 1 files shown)


jeanconn:
Regarding "should we document here that slight changes to COLOR1 to avoid 0.7 and 1.5?" - I'm thinking that's a tiny processing detail that maybe doesn't make the final public doc for COLOR1 (which currently has)
```
COLOR1 - float variable expressing the cataloged or estimated B-V color,
        used for mag_aca, in mag.  If no colors are available, the default
        value is 0.7000.  If Tycho-2 data show colors redder than
        (B-V)=1.5, then the default value is 1.5000.  True cataloged
        color values are stored in COLOR2.
```
The details about 0.7000 and 1.5000 remain true.

javierggt:
This would have been good to know. Sorry I didn’t notice. A proper catalog update would have had a calibration of the Gaia color, although I doubt that a simple scaling factor would have been enough.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

jeanconn:
Though that kinda raises the question - should we make the agasc-gaia repo public and link to *it* from the README?

taldcroft:
> Though that kinda raises the question - should we make the agasc-gaia repo public and link to _it_ from the README?

I would prefer not since there is a lot of uncurated material there.

taldcroft:
About:
```
 If Tycho-2 data show colors redder than (B-V)=1.5, then the default value is 1.5000.
```
That is not precisely correct if the AGASC catalog color is now from Gaia but Tycho-2 shows B-V > 1.5 . I would suggest:
```
If the color is derived from Tycho-2 (C1_CATID=5) and that color is redder than (B-V)=1.5 then
COLOR1 is set to 1.5000. This is the case for about 21,000 stars in AGASC 1.8. True cataloged 
color values are stored in COLOR2.
```

taldcroft:
About the `alpha` issue, I think we don't care and that leaving it blank as in this PR is fine.

taldcroft:
And I didn't see the `I figure we should just ignore alpha in the context of the AGASC 1.8 readme update.` bit. So we are in agreement.

taldcroft:
Apart from my one comment, I think this is good enough.

jeanconn:
OK I updated the COLOR1 text per Tom and a tiny typo while I was in the neighborhood.

taldcroft (approved):

PR #190, opened 2024-08-28
Get last_updated supplement table without error

javierggt:
## Description

This PR fixes an exception that happens when one does `agasc.get_supplement_table("last_updated")` without specifying `as_dict=True`. Specifying `as_dict=True` is not intuitive and not expected, and anyway 

### Unit tests

Added a unit test for this and checked that it fails on master as expected.

- [ ] No unit tests
- [x] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests

No functional testing.

### DIFF
--- a/agasc/tests/test_agasc_2.py
+++ b/agasc/tests/test_agasc_2.py
@@ -26,12 +26,13 @@ from pathlib import Path
 
 import numpy as np
 import pytest
-import Ska.Shell
 import tables
 from astropy.io import ascii
 from astropy.table import Row, Table
+from cxotime import CxoTime
 
 import agasc
+import Ska.Shell
 from agasc import write_agasc
 
 os.environ[agasc.SUPPLEMENT_ENABLED_ENV] = "False"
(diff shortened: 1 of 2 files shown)


agasc/tests/test_agasc_2.py
```diff
@@ -641,3 +642,10 @@ def test_write_missing_column(tmp_path, stars_in):
 
     # OK for full_agasc=False
     write_agasc(temp, stars=stars, version="test", full_agasc=False)
+
+
+def test_last_updated():
+    last_updated = agasc.get_supplement_table("last_updated")
+    assert CxoTime(last_updated["supplement"]).date.startswith("20")
```
taldcroft, line 649: I see that there is no other testing of these supplement tables. So while you are at it, it would be worth testing:
- `last_updated` has the correct four keys
- Each corresponding value is a string that is a valid ISO format date `CxoTime(last_updated[key], format="iso")`.
- `agasc_versions` has the same four keys.

The name of the test function does not quite match what it does.

taldcroft (changes requested):
Looks good. Please always document your testing with the `git rev-parse HEAD` output and a copy/paste of the `pytest` results along with any custom setup. It's a slight hassle but otherwise there is no record of which version was tested.

taldcroft:
@javierggt - is this PR overtaken  by #203 or events?

javierggt:
> @javierggt - is this PR overtaken by #203 or events?

No. It's still valid. I added some text to the functional test description so you can check.

My guess is that this PR stalled after you asked for further changes. I made those changes now, so this should be good.

agasc/tests/test_agasc_2.py
javierggt: I think I addressed this comment in my last commit.

PR #194, opened 2024-11-18
Add version 1.8 files of the text files stored with the fits files

jeanconn:
## Description

Add version 1.8 files of the text files stored with the fits files

### Unit tests

- [x] Skipped in this docs case

### DIFF
--- a/docs/comments.txt
+++ b/docs/comments.txt
@@ -1,25 +1,25 @@
 
 
-            THE AXAF GUIDE and ACQUISITION STAR CATALOG V1.7
-                                                                
-             An all-sky astrometric and photometric catalog     
-                 prepared for the operation of Chandra 
-                 (formerly AXAF). version 1.7 September 2018
+            THE AXAF GUIDE and ACQUISITION STAR CATALOG V1.8
+
+             An all-sky astrometric and photometric catalog
+                 prepared for the operation of Chandra
+                 (formerly AXAF). version 1.8 July 2024
 
 
 
                              File COMMENTS
-                   Introduction and General Comments           
-                                                                
+                   Introduction and General Comments
+
 This table contains text (in ASCII 80 column format) summarizing the
-organization, format, and data in the AGASC1.7.  The content and data
-sources for the AGASC1.7 are also described.
+organization, format, and data in the AGASC1.8.  The content and data
+sources for the AGASC1.8 are also described.
 
 CONTENTS:
 	1. INTRODUCTION
 	2. ORGANIZATION OF THE DATA FILES
-	3. SUMMARY OF THE AGASC1.7 FORMAT
-	4. HISTORY OF THE AGASC version 1.7
+	3. SUMMARY OF THE AGASC1.8 FORMAT
+	4. HISTORY OF THE AGASC version 1.8
 		4.1 HST GSC
 		4.2 POSITIONS AND PROPER MOTIONS CATALOG (PPM)
 		4.3 TYCHO OUTPUT CATALOG (TYC)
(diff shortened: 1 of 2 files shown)


jeanconn:
Hi @javierggt - I have an ancient open action to update /data/agasc1p8/readme.txt and /data/agasc1p8/tables/comments.txt with 1.8 versions.  I think they should just be the updated versions of the 1.7 files (found in the expected locations in /data/agasc1p7) and not just copies of the catalog description rst.  So perhaps take a look at https://github.com/sot/agasc/pull/194/commits/34d634daf73fad843f50ffebd45a00d279938571 in here and see if those changes seem reasonable to you?

javierggt:
I disagree.

I think whatever is in the rst should be the same as the readme. Already the fact that there are two files (readme an comments.txt) makes it harder to know where things should be documented. With this, we would be increasing the number of files rather than decreasing it. Note that there are already two RST files!

In general, there should be only one authoritative source of information.

If anything, we should reduce the number of files to one.

jeanconn:
OK, then go ahead make the txt files as 80 column files to replace the ones that are there?  It looked to me like the description rst did not match the readme.txt or the comments.txt file in content and it seemed like the minimal process to update the files that are expected.  Which two rst files are you talking about?

docs/comments.txt
```diff
@@ -0,0 +1,813 @@
...
+2    CLASS - short integer code identifying classification of entry.
+     Default value of 0.
+        0 - star
+        1 - galaxy
+        2 - blend or member of incorrectly resolved blend.
+        3 - non-star
+        5 - potential artifact
+        6 - known multiple system
+        7 - close to galaxy or other extended object
+
```
jeanconn, line 255: I note the general agasc description rst now talks about the class=100+ cases, but that doesn't apply to the fits files.

jeanconn: I hear you about documentation for *us*.  But the previously existing system, and the one that is actually documented in the catalog-description.rst, points to having a readme.txt at top level and a comments.txt file in the tables directory.  If you want to write a script to make those files from catalog-description.rst that's one option.  Another option was just to make them manually, which I did here with the idea that really nobody uses the catalog except us anyway.  A third option would be to update the catalog-description.rst to remove the references to readme.txt and comments.txt and replace the readme with a text file that just says "See https://sot.github.io/agasc/catalog-description.html" .

jeanconn: Oh, and catalog-description-1p7.rst was needed so that both could actually be available from our documentation at https://sot.github.io/agasc/ .  At the time, both agasc versions were actually in use.  If there's a driver to remove the previous documentation we can talk about that as well.

jeanconn:
@javierggt , in the end, before the next POG is released, I just want to ask arcops to update https://cxc.harvard.edu/cgi-gen/cda/agasc/agascInterface.pl to point to the 1.8 files and update the links to "Detailed Information" to point to something reasonable.  My first thought was to just give them updated files matching what was already there (this PR), but if that really bothers you, what would you like those "Detailed Information" links to point to?  sot.github.com/agasc ?

javierggt:
My questions still stand. 

Considering:
- the rst file was first created by copying the 1p7 readme
- I added information to the rst file in PR #182 and I totally thought we would copy that to the file directory.
- The diffs make it hard to see exactly what changed. I can read the files, but I might miss something.

My questions:
- Is all the info currently in the RST file included in this PR's new files?
- Do the files in this PR include information that is not in the RST file?
- Other than just removing the lines with CATID=100, is there a reason for duplicating and going back to an older file?

I'm arguing for something that I thought would be uncontroversial: we should not duplicate gratuitously, there should be a single source of information, and whatever we say the source is, it should be improved, rather than creating more versions.

The RST file already is a copy of the readme with improvements.

javierggt:
> I just want to ask arcops to update https://cxc.harvard.edu/cgi-gen/cda/agasc/agascInterface.pl to point to the 1.8 files and update the links to "Detailed Information" to point to something reasonable.

Why is the current RST file "not reasonable"?

jeanconn:
> Is all the info currently in the RST file included in this PR's new files?

I do not know.  My goal with the PR was just to update the readme.txt and the comments.txt so they had information that matched what was in the the RST file at all point where there was a difference.  I did not confirm that all information in the RST file has a home in the text files.

jeanconn:
> Do the files in this PR include information that is not in the RST file?

The readme and the comments file have a different organization than the RST file but I agree that the RST file seems to have everything that is in the two text files.

jeanconn:
> Other than just removing the lines with CATID=100, is there a reason for duplicating and going back to an older file?

I don't know if any external resources use the locations of readme.txt and comments.txt.  That's basically why I thought that updating those files made the most sense to just not worry about this.   It looks like the link from the CGI page https://cxc.harvard.edu/cgi-gen/cda/agasc/agascInterface.pl points to https://cxc.harvard.edu/cda/agasc/agasc1p7_comments.txt so it is probably a copy anyway as opposed to a link into /data/agasc1p7/tables/comments.txt .

jeanconn:
> Why is the current RST file "not reasonable"?

Where do you want to link to it?  I also wasn't as confident about RST rendering or just displaying as text in all browsers, but don't care.

jeanconn:
I'll just send the reference to the RST file to arcops and they can figure out someplace to put it.

PR #195, opened 2024-11-19
Update for public use.

jeanconn:
## Description

This PR:

- Removes references to text files no longer supported.
- Updates the text to explain what the agasc package is.

### Unit tests

Docs only update - tests not run

### DIFF
--- a/docs/catalog-description.rst
+++ b/docs/catalog-description.rst
@@ -69,6 +69,17 @@ Version 1.8 includes improved star positions and proper motions from Gaia
 DR3, as well as improved estimated magnitudes derived from Gaia DR3 photometry
 and more than 94,000 ACA observed stars. 
 
+
+Use of the AGASC 1.8 in Chandra Operations
+==========================================
+
+The Chandra Aspect Team uses an HDF5 version of the AGASC 1.8 data and
+the Python agasc package https://sot.github.io/agasc/index.html. References
+in this document to the agasc package refer to that Python code.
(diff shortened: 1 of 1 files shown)


jeanconn:
Though if we're going to include the catalog-description.rst as the public docs for AGASC 1.8, it probably needs a link within it to the github docs to explain what an "agasc package query" is.

docs/catalog-description.rst
```diff
@@ -69,21 +69,30 @@ Version 1.8 includes improved star positions and proper motions from Gaia
...
+directly with that Python package.
+
+
 Organization of the FITS Data Files
 ===================================
 
 The top directory includes the files::
 
     ----------------------------------------------------------------
-    readme.txt    - Introduction. ASCII
```
jeanconn, line 78: Question here if we should add catalog-description.rst to the file list or not bother.

javierggt: I was thinking this file would be renamed readme.txt when placing it with the fits file distribution.

The `readme.txt` could be replace by `this file`

javierggt: either way, just removing the line is ok for me.

jeanconn: OK, if you were thinking that it was fine to drop the ".rst" extension and just call catalog-description.rst the new readme.txt that's fine too.

jeanconn: Done.

jeanconn:
I'm planning to ignore unrelated ruff issues.

javierggt (approved):

PR #199, opened 2025-06-25
Zero size error

javierggt:
## Description

This PR makes changes so that the following error found recently is properly reported:
```
Unexpected Error: zero-size array to reduction operation minimum which has no identity
```

The specific occurrence of this error is due to missing telemetry. I did not think this could happen, because this happens right after the output from `aca_l0.get_slot_data` is checked. I don't see how one can have slot data and no telemetry.

Actually, it can happen if there is telemetry but `fetch.MSIDset.filter_bad` removes everything, because the telemetry is filtered before checking. I re-ran the script and didn't reproduce the error, so I do not think it is due to filtering. I figure we at least want to label this error from now on.

This actual changes in this PR:
- raise an exception for no telemetry (after filtering)
- updated the "known" exceptions
   - removed messages that are never issued and included "No matching times between cheta and level0" (this message changed and it was never updated)
- fixed a bug where the OBSID was being wrongly reported (it was reporting the AGASC ID instead)

### Unit tests

- [ ] No unit tests
- [ ] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests

No functional testing.

### DIFF
--- a/agasc/supplement/magnitudes/mag_estimate.py
+++ b/agasc/supplement/magnitudes/mag_estimate.py
@@ -373,6 +373,14 @@ def get_telemetry(obs):
             slot=obs["slot"],
         )
     times = msids[f"AOACMAG{slot}"].times
+    if len(times) == 0:
+        raise MagStatsException(
+            "No telemetry data",
+            agasc_id=obs["agasc_id"],
+            obsid=obs["obsid"],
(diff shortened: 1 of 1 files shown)


taldcroft (approved):
Looks fine.

PR #208, opened 2026-03-20
Refactor observation statistics processing and discard recent observations without telemetry

javierggt:
## Description

This PR refactors the way obs-stats are calculated to disentangle it from the agasc_stats (the supplement is based on stars, so the processing did not allow direct processing of observations independently).

After that, I fixed a few issues:
- _Recent_ observations without telemetry are discarded before trying to process them. With this change, recent observations without telemetry should not show up in any warning in any report. Older observations without telemetry are still shown in yellow (as can be seen in functional testing)
- Added descriptions for the sections in the supplement report.

This PR fixes a few other issues:
- When there is an unexpected error, now the error message is reported instead of showing "Unknown error". This can be seen in the functional test: some stars have `PermissionError`
- By dropping the newer observations without telemetry, the last_observed date is before the dropped observation, and this protects against the case where there is data for a later observation, which would set last_observed to a later date, causing the observation to be skipped on subsequent runs.

### Unit tests

- [x] Linux

```
jgonzale agasc $ git rev-parse HEAD
daf4a4a32fbe99b19e5abfa3f19bf0f9ca8caece
jgonzale agasc $ pytest agasc
========================================================= test session starts =========================================================
platform linux -- Python 3.13.11, pytest-9.0.2, pluggy-1.6.0
rootdir: /proj/sot/ska/jgonzalez/git
configfile: pytest.ini
plugins: anyio-4.12.1, timeout-2.4.0
collected 79 items                                                                                                                    

agasc/tests/test_agasc_1.py .......                                                                                             [  8%]
agasc/tests/test_agasc_2.py ...............................................                                                     [ 68%]
agasc/tests/test_agasc_healpix.py ...........                                                                                   [ 82%]
agasc/tests/test_obs_status.py ..............                                                                                   [100%]

=================================================== 79 passed in 105.09s (0:01:45) ====================================================

```

- [x] Linux
```
jeanconn-kady> pytest
============================================ test session starts =============================================
platform linux -- Python 3.13.11, pytest-9.0.2, pluggy-1.6.0
rootdir: /proj/sot/ska/jeanproj/git
configfile: pytest.ini
plugins: anyio-4.12.1, timeout-2.4.0
collected 79 items                                                                                           

agasc/tests/test_agasc_1.py .......                                                                    [  8%]
agasc/tests/test_agasc_2.py ...............................................                            [ 68%]
agasc/tests/test_agasc_healpix.py ...........                                                          [ 82%]
agasc/tests/test_obs_status.py ..............                                                          [100%]

======================================= 79 passed in 116.16s (0:01:56) =======================================
jeanconn-kady> git rev-parse HEAD
daf4a4a32fbe99b19e5abfa3f19bf0f9ca8caece
```
### Functional tests

Running weekly update in test environment on `2026:187`. Output:
- [flight](https://cxc.cfa.harvard.edu/mta/ASPECT/agasc/supplement_reports/weekly/2026:187)
- [branch](https://icxc.cfa.harvard.edu/aspect/test_review_outputs/agasc/pr-208/supplement_reports/weekly/2026:187)

The differences summarized:
- There are two stars in flight and not in the branch (31069328 and 897587816) because they are right at the edge of the time range. These two stars are scheduled in OBSID 30267 and 31346. The last observations in flight are [32412, 31404, 31346, 30267] whereas the branch stopped at 32412. For these stars, there are no observations within the time range considered after you remove these three OBSIDs. Other stars in these three OBSIDs do have observations in the time range.
- 23 stars are in flight and not in the branch because all their observations have "no telemetry" errors, so the branch treats them as if they do not exist yet.
- 50 stars have observations with "no telemetry" errors in flight which got dropped in the branch. For all these, we checked that the last observation date is earlier than the dropped observation (no instances where there is a later observation). Of these:
  - 13 show as "magnitude not updated" (there are _other_ new observations for processing, there are mag updates, but the mag changes are small -> no update)
  - 37 show as "in limbo". These are the ones where there are no new observations up for processing (they are processed already) so nothing was done even though they were in the list. Note the "suspect" observation also in limbo (that just means there was no mag update even though it was in the list.
- AGASC ID 31983336 had file permission issues

### DIFF
--- a/agasc/supplement/magnitudes/mag_estimate.py
+++ b/agasc/supplement/magnitudes/mag_estimate.py
@@ -516,6 +516,13 @@ def get_telemetry_by_agasc_id(
     )
 
 
+def _has_error(telem):
+    # error entries produced by get_telemetry_by_observations are plain dicts;
+    # successful telemetry can be a dict or an astropy Table, and
+    # `"error_code" in <Table>` raises TypeError under numpy >= 2
+    return isinstance(telem, dict) and "error_code" in telem
+
+
 def get_telemetry_by_observations(observations, ignore_exceptions=False, as_table=True):
     telem = []
     for obs in observations:
@@ -560,7 +567,7 @@ def get_telemetry_by_observations(observations, ignore_exceptions=False, as_tabl
         # we ignore entries with error_code here because this means ignore_exceptions is true
         # (otherwise the exception would have been raised already)
         return (
-            vstack([Table(tel) for tel in telem if "error_code" not in tel])
+            vstack([Table(tel) for tel in telem if not _has_error(tel)])
             if as_table
             else telem
         )
@@ -1233,8 +1240,7 @@ def get_agasc_id_stats(
     # discard observations with no telemetry if they occured in the last two weeks.
     discard = np.array(
         [
-            ("error_code" in tlm)
-            and (tlm["mp_starcat_time"] > CxoTime() - no_telem_time)
+            _has_error(tlm) and (tlm["mp_starcat_time"] > CxoTime() - no_telem_time)
             for tlm in all_telem
         ]
     )
@@ -1649,7 +1655,7 @@ def get_multi_obs_stats(star_obs, telem=None, obs_status_override=None):
             comment = status["comments"]
 
         obs_telem = telem[i]
-        if "error_code" in obs_telem:
+        if _has_error(obs_telem):
             fail = obs_telem["error_code"] > 2
             obs_stat = get_obs_stats(obs, telem=[])
             obs_stat.update(
--- a/agasc/supplement/magnitudes/mag_estimate.py
+++ b/agasc/supplement/magnitudes/mag_estimate.py
@@ -1258,6 +1258,10 @@ def get_agasc_id_stats(
         stats["w"] = np.array([])
         stats["mean_corrected"] = np.array([])
         stats["weighted_mean"] = np.array([])
+        weighted_mean = {
+            "mag_weighted_mean": [],
+            "mag_weighted_std": [],
+        }
 
     star = get_star(agasc_id, use_supplement=False)
 
--- a/agasc/supplement/magnitudes/mag_estimate.py
+++ b/agasc/supplement/magnitudes/mag_estimate.py
@@ -481,7 +481,9 @@ def get_telemetry(obs):
     return telem
 
 
-def get_telemetry_by_agasc_id(agasc_id, obsid=None, ignore_exceptions=False):
+def get_telemetry_by_agasc_id(
+    agasc_id, obsid=None, ignore_exceptions=False, as_table=True
+):
     """
     Get all telemetry relevant for the magnitude estimation, given an AGASC ID.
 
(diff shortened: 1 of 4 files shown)


taldcroft:
What's happening on this?

javierggt:
If you check the output you will see that this PR is ok as far as labeling "no-telemetry" observations, but every once in a while there are observations that fail with an unspecified error (like [this](https://icxc.cfa.harvard.edu/aspect/test_review_outputs/agasc/pr-208/supplement_reports/weekly/2026:138/stars/003/31983336/index.html)). On Wednesday I tried, and got no error processing those failed observations, so I don't know why they failed. I added some debug output to check on Monday.

agasc/supplement/magnitudes/mag_estimate.py
```diff
@@ -491,10 +494,13 @@ def get_telemetry_by_agasc_id(agasc_id, obsid=None, ignore_exceptions=False):
...
     :param ignore_exceptions: bool
         if True, any exception is ignored. Useful in some cases. Default is False.
+    :param as_table: bool
+        if True, the telemetry is returned as an astropy Table. If False, it is returned as a list
+        of dicts, one per observation. Default is True.
     :return: dict
     """
     logger.debug(f"  Getting telemetry for AGASC ID={agasc_id}")
-    star_obs_catalogs.load()
+    star_obs_catalogs.load()  # should this be here?
```
jeanconn, line 503: The comment doesn't inspire confidence but...

jeanconn:
This looks good to me overall.  One thing I note is that with the transition from "Other (unexpectedly not updated) " to "Stars in Limbo " the PR (and previous ones I think) also filter out all of the information that one might use to see why the star is actually in limbo.  So I'm not sure if that section of the report actually helps us?  But it probably makes sense to keep an eye on that and then see if it can get cut altogether in a future update.

jeanconn (approved):

javierggt:
What do you mean by "the PR (and previous ones I think) also filter out all of the information that one might use to see why the star is actually in limbo"? What information is that?

Nothing is new being filter out, except dropping the observations with "no telemetry" error. Is that what you mean? Other than that, the section was just renamed and nothing changed.

jeanconn:
Right, so yes - I can't tell that the stars are in limbo because of the no-telemetry issue because the no-telemetry stuff doesn't get logged.  I couldn't remember if there are other reasons they could be in limbo - again I couldn't find anything in the limbo-stars reports to tell me why they are there.

javierggt:
That section is a list of stars that "should" have been updated, but at the end of processing there is no update. In other words: there is a function that takes a list of stars as input and gives you an update for each. If a star is in the list but there is no result, it ends up in this list. That is the exact meaning.

The intention from the beginning was to be like an accountant: everything that goes in must come out.

I would appreciate it if you or Tom propose a description so we do not have to go over this again, because I have tried multiple times and it never seems to be enough. The current description tries to say what I just said and even lists two possibilities for why a star is in this list: no telemetry because observations are too new, or suspect observations.

This PR does exactly what Tom has requested: if an observation has no telemetry, drop it on the floor. Nothing gets logged because it was dropped.

jeanconn:
Yeah, I was not casting shade here and I think this PR is fine.  I think we can also probably decide together if, going forward, the "limbo" section should stay as it is, get more debug information, or get cut.  

I'm trying to decide now if, when I'm on ACA review, I can just safely ignore the section or if I should pay more attention to it if it gets above a certain number of stars.

jeanconn (changes requested):
Though actually I just realized that in this branch tests don't pass for me.

javierggt:
@jeanconn I made two commits just now:
- fix the test
- remove that suspicious line where I wrote "maybe this is not needed"

I admit that, when I wrote that line I did not think much, and I must have been thinking something like "why is this even needed?". I committed it with a WIP comment to remind me to go back to it.

So... now the tests pass, but maybe we want to leave this PR open until this last commit runs on Sunday.

jeanconn:
Sounds good.  I'll just take an action to review the test output after the next run.

javierggt:
This week's output looks OK. This week's output is not as interesting as last week's (which I linked in the description) because it doesn't have all the kinds of failures.

Anyway, the last changes would not change the output, so just the fact that it ran without issues is good by me.

jeanconn:
Sounds good - I'll run unit tests again for the description in record - would you please do the same to close this out?

javierggt:
Done

jeanconn (approved):
Thanks!

